PR #4, opened 2024-06-20
Use and configure nginx

mmkay:
This PR configures nginx in a way so that all instances of tempo-worker are accessible on all possible ports and forward both the HTTP and GRPC traffic to correct upstreams.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -89,8 +89,9 @@ class TempoCoordinatorCharm(CharmBase):
             ],
         )
         # # Patch the juju-created Kubernetes service to contain the right ports
-        external_ports = tempo.get_external_ports(self.app.name)
-        self._service_patcher = KubernetesServicePatch(self, external_ports)
+        self.unit.set_ports(*self.tempo.all_ports.values())
+
+        # self._service_patcher = KubernetesServicePatch(self, external_ports)
         # Provide ability for Tempo to be scraped by Prometheus using prometheus_scrape
         self._scraping = MetricsEndpointProvider(
             self,
--- a/src/charm.py
+++ b/src/charm.py
@@ -292,8 +292,14 @@ class TempoCoordinatorCharm(CharmBase):
     def _on_cert_handler_changed(self, _):
         if self.tls_available:
             logger.debug("enabling TLS")
+            self.nginx.configure_tls(
+                server_cert=self.cert_handler.server_cert,  # type: ignore
+                ca_cert=self.cert_handler.ca_cert,  # type: ignore
+                private_key=self.cert_handler.private_key,  # type: ignore
+            )
         else:
             logger.debug("disabling TLS")
+            self.nginx.delete_certificates()
 
         # tls readiness change means config change.
         # sync scheme change with traefik and related consumers
@@ -389,7 +395,7 @@ class TempoCoordinatorCharm(CharmBase):
                 e.add_status(ActiveStatus())
 
     def _on_nginx_pebble_ready(self, _) -> None:
-        self.nginx.configure_pebble_layer(tls=self.tls_available)
+        self.nginx.configure_pebble_layer()
 
     def _on_nginx_prometheus_exporter_pebble_ready(self, _) -> None:
         self.nginx_prometheus_exporter.configure_pebble_layer()
@@ -399,7 +405,7 @@ class TempoCoordinatorCharm(CharmBase):
         if isinstance(event, CollectStatusEvent):
             return
         # plan layers
-        self.nginx.configure_pebble_layer(tls=self.tls_available)
+        self.nginx.configure_pebble_layer()
         self.nginx_prometheus_exporter.configure_pebble_layer()
         # configure ingress
         self._configure_ingress()
--- a/src/charm.py
+++ b/src/charm.py
@@ -157,7 +157,7 @@ class TempoCoordinatorCharm(CharmBase):
         self.framework.observe(self.tempo_cluster.on.changed, self._on_tempo_cluster_changed)
 
         for evt in self.on.events().values():
-            self.framework.observe(evt, self._on_event)
+            self.framework.observe(evt, self._on_event)  # type: ignore
 
     ######################
     # UTILITY PROPERTIES #
@@ -394,7 +394,7 @@ class TempoCoordinatorCharm(CharmBase):
     def _on_nginx_prometheus_exporter_pebble_ready(self, _) -> None:
         self.nginx_prometheus_exporter.configure_pebble_layer()
 
-    def _on_event(self, event):
+    def _on_event(self, event) -> None:
         """A set of common configuration actions that should happen on every event."""
         if isinstance(event, CollectStatusEvent):
             return
@@ -426,7 +426,7 @@ class TempoCoordinatorCharm(CharmBase):
         # notify the cluster
         self._update_tempo_cluster()
 
-    def _update_tracing_relations(self):
+    def _update_tracing_relations(self) -> None:
         tracing_relations = self.model.relations["tracing"]
         if not tracing_relations:
             # todo: set waiting status and configure tempo to run without receivers if possible,
@@ -454,12 +454,12 @@ class TempoCoordinatorCharm(CharmBase):
         requested_receivers = requested_protocols.intersection(set(self.tempo.receiver_ports))
         return tuple(requested_receivers)
 
-    def server_cert(self):
+    def server_cert(self) -> str:
         """For charm tracing."""
         self._update_server_cert()
         return self.tempo.server_cert_path
 
-    def _update_server_cert(self):
+    def _update_server_cert(self) -> None:
         """Server certificate for charm tracing tls, if tls is enabled."""
         server_cert = Path(self.tempo.server_cert_path)
         if self.tls_available:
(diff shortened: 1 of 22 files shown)


src/charm.py
```diff
@@ -368,6 +388,26 @@ def _on_collect_unit_status(self, e: CollectStatusEvent):
             else:
                 e.add_status(ActiveStatus())
 
+    def _on_nginx_pebble_ready(self, _) -> None:
+        self.nginx.configure_pebble_layer(tls=self.tls_available)
```
michaeldmitry, line 392: instead of propagating `self.tls_available` down to where its needed, can we make, for example, a new property in `nginx.py` `tls_ready` that checks if certs are already present in the disk or not?

src/charm.py
```diff
@@ -368,6 +388,26 @@ def _on_collect_unit_status(self, e: CollectStatusEvent):
...
+        """A set of common configuration actions that should happen on every event."""
+        if isinstance(event, CollectStatusEvent):
+            return
+        # plan layers
+        self.nginx.configure_pebble_layer(tls=self.tls_available)
+        self.nginx_prometheus_exporter.configure_pebble_layer()
+        # configure ingress
+        self._configure_ingress()
+        # update cluster relations
+        self._update_tempo_cluster()
```
michaeldmitry, line 407: I think this causes that a new tempo config gets pushed to the workers. The workers then reconcile and every time this happens, tempo will be down until the new config is reflected and tempo is up again. It causes unneeded `cluster-changed` event at least every 5 mins (with update-status event)

src/nginx.py
```diff
@@ -297,35 +166,77 @@ def _log_verbose(self, verbose: bool = True) -> List[Dict[str, Any]]:
...
-            nginx_locations.extend(LOCATIONS_COMPACTOR)
+    def _upstream(self, role: str, address_set: Set[str], port: int) -> Dict[str, Any]:
+        return {
+            "directive": "upstream",
+            "args": [role],
+            "block": [{"directive": "server", "args": [f"{addr}:{port}"]} for addr in address_set],
+        }
+
+    def _locations(self, upstream: str, grpc: bool) -> List[Dict[str, Any]]:
+        protocol = "grpc" if grpc else "http"
```
michaeldmitry, line 227: shouldn't here we have sth like `protocol = "grpc" if grpc else ("https" if self.tls_ready else "http")`
When we enable TLS, workloads will also be run using TLS. So, internal communication using what's here (http) will fail as its expecting https.

src/nginx.py
```diff
@@ -344,16 +255,63 @@ def _basic_auth(self, enabled: bool) -> List[Optional[Dict[str, Any]]]:
...
+            args.append("http2")
+        return args
+
+    def _servers(
+        self, addresses_by_role: Dict[str, Set[str]], tls: bool = False
+    ) -> List[Dict[str, Any]]:
+        servers = []
+        roles = addresses_by_role.keys()
+
+        if "distributor" in roles or "all" in roles:
```
michaeldmitry, line 286: can we extract those literals as constants somewhere

src/tempo.py
```diff
@@ -207,7 +207,9 @@ def _build_storage_config(self, s3_config: dict):
                 endpoint=s3_config["endpoint"],
                 secret_key=s3_config["secret-key"],
             ),
-            block=tempo_config.Block(version="v2"),
+            # starting from Tempo 2.4, we need to use at least parquet v3 to have search capabilities (Grafana support)
+            # https://grafana.com/docs/tempo/latest/release-notes/v2-4/#vparquet3-is-now-the-default-block-format
+            block=tempo_config.Block(version="vParquet3"),
```
michaeldmitry, line 212: Tempo will now give a warning that property `v2_out_buffer_bytes` in line `255` can be used when version is set to "v2". Should we worry about this?

src/charm.py
```diff
@@ -368,6 +388,26 @@ def _on_collect_unit_status(self, e: CollectStatusEvent):
             else:
                 e.add_status(ActiveStatus())
 
+    def _on_nginx_pebble_ready(self, _) -> None:
```
michaeldmitry, line 391: When TLS is enabled, shouldn't we push certs into the container of nginx?

michaeldmitry (commented):
Nice work! Few comments :)

src/charm.py
mmkay: True, that looks much better :)

src/charm.py
mmkay: Not sure as I think the assumption was that it won't propagate to the workers if everything is the same: https://github.com/canonical/tempo-coordinator-k8s-operator/blob/main/src/charm.py#L493

Will test it out further tomorrow.

src/nginx.py
mmkay: True, and also `grpcs` for SSL-enabled setup. Added.

src/tempo.py
mmkay: Removed the property altogether.

src/charm.py
mmkay: Absolutely, good catch :) Thanks to that we were actually doing unsecured communication with workers, it seems. Updated.

src/charm.py
mmkay: Doesn't seem like I'm getting the extra events?

```
ubuntu@charm-dev:~/tempo-coordinator-k8s-operator$ juju show-status-log tempo-worker-k8s/0
Time                        Type       Status     Message
10 Jul 2024 18:46:28+02:00  juju-unit  error      hook failed: "start"
10 Jul 2024 18:46:33+02:00  juju-unit  executing  running start hook
10 Jul 2024 18:46:36+02:00  juju-unit  executing  running tempo-pebble-ready hook
10 Jul 2024 18:47:40+02:00  juju-unit  idle       
10 Jul 2024 19:02:58+02:00  juju-unit  executing  running tempo-cluster-relation-changed hook
10 Jul 2024 19:03:10+02:00  juju-unit  idle       
10 Jul 2024 19:20:55+02:00  juju-unit  executing  running tempo-cluster-relation-departed hook for tempo-coordinator-k8s/0
10 Jul 2024 19:20:56+02:00  workload   active     (all roles) ready.
10 Jul 2024 19:20:56+02:00  juju-unit  executing  running tempo-cluster-relation-broken hook
10 Jul 2024 19:20:57+02:00  workload   blocked    Missing tempo-cluster relation to a tempo-coordinator charm
10 Jul 2024 19:20:57+02:00  juju-unit  idle       
10 Jul 2024 19:22:39+02:00  juju-unit  executing  running tempo-cluster-relation-created hook
10 Jul 2024 19:22:40+02:00  juju-unit  idle       
10 Jul 2024 19:23:05+02:00  juju-unit  executing  running tempo-cluster-relation-joined hook for tempo-coordinator-k8s/0
10 Jul 2024 19:23:06+02:00  juju-unit  executing  running tempo-cluster-relation-changed hook for tempo-coordinator-k8s/0
10 Jul 2024 19:23:06+02:00  workload   waiting    Waiting for coordinator to publish a tempo config
10 Jul 2024 19:23:06+02:00  juju-unit  idle       
10 Jul 2024 19:23:16+02:00  juju-unit  executing  running tempo-cluster-relation-changed hook
10 Jul 2024 19:23:32+02:00  juju-unit  idle       
11 Jul 2024 11:45:24+02:00  workload   active     (all roles) ready.
```

src/nginx.py
mmkay: Simplified the code significantly by using values from enums/dicts from `Tempo` where it was possible. Now it should look better :)

src/charm.py
michaeldmitry: maybe its sth in my setup.

tests/scenario/test_nginx.py
```diff
@@ -0,0 +1,98 @@
...
+            "distributor": {"1.2.3.5", "1.2.3.7"},
+            "ingester": {"1.2.3.6", "1.2.3.8"},
+            "querier": {"1.2.4.7", "1.2.4.9"},
+            "query_frontend": {"1.2.5.1", "1.2.5.2"},
+            "compactor": {"1.2.6.6", "1.2.6.7"},
+            "metrics_generator": {"1.2.8.4", "1.2.8.5"},
+        },
+    ),
+)
+def test_nginx_config_is_parsed_with_workers(
```
michaeldmitry, line 85: is it possible to add a test to check on pieces of the generated config. The generation logic does lots of things and it feels like a test for the actual content would be beneficial.

mmkay: Yes, I added checking whether expressions like `listen`, `grpc_pass`, `proxy_pass`, `upstream` are generated with correct ports and names. Good idea :)

src/charm.py
```diff
@@ -79,8 +88,9 @@ def __init__(self, *args):
             ],
         )
         # # Patch the juju-created Kubernetes service to contain the right ports
-        external_ports = tempo.get_external_ports(self.app.name)
-        self._service_patcher = KubernetesServicePatch(self, external_ports)
+        self.unit.set_ports(*self.tempo.all_ports.values())
+
+        # self._service_patcher = KubernetesServicePatch(self, external_ports)
```
michaeldmitry, line 93: Do we need this comment

mmkay: Seems like not anymore.

michaeldmitry (approved):

PR #5, opened 2024-06-21
translate tempo config to pydantic models

michaeldmitry:
## Issue
tempo configurations are in the form of raw dicts.

## Solution
Translate all currently used tempo config to pydantic models.

## Testing instructions
For coordinator to work and push tempo configuration to workers, it needs an s3 integration. To workaround the issue and not need an actual s3 instance, we can use Facade charm for mocking s3 interface.

### Deploy tempo k8s operator (used for validation)
```
juju deploy tempo-k8s tempo --resource tempo-image=grafana/tempo:2.4.0 --channel=latest/edge
```

### Deploy tempo coordinator
cwd to `tempo-coordinator-k8s-operator`

```
charmcraft pack
juju deploy ./tempo-coordinator-k8s_ubuntu-22.04-amd64.charm 
```

### Deploy Facade
cwd to `tempo-coordinator-k8s-operator`
```
juju deploy facade s3-facade --channel=latest/edge
juju integrate s3-facade:provide-s3 tempo-coordinator-k8s:s3
juju run s3-facade/0 update --params ./tests/manual/facades/s3.yaml
```

### Deploy tempo worker
cwd to `tempo-worker-k8s-operator`

```
git fetch
git checkout coordinator-integration
charmcraft pack
juju deploy ./tempo-worker-k8s_ubuntu-22.04-amd64.charm  tempo-worker --resource tempo-image=docker.io/ubuntu/tempo:2-22.04
```
### Integrate cluster
```
 juju integrate tempo-coordinator-k8s tempo-worker
```

### Validation

1- coordinator and worker should both be in active/idle
2- kubectl exec -it tempo-worker-0 -n <model> -c tempo -- cat /etc/tempo/tempo.yaml
3- kubectl exec -it tempo-0 -n <model> -c tempo -- cat /etc/tempo/tempo.yaml
4- Both config files should match apart from storage options, as s3 would be used in coordinator instead of local storage

### TLS Validation
Repeat above steps after ` juju integrate ssc tempo-coordinator-k8s`

### DIFF
--- a/src/tempo.py
+++ b/src/tempo.py
@@ -187,12 +187,12 @@ class Tempo:
             )
             config.memberlist = config.memberlist.model_copy(update=tls_config)
 
-        return config.model_dump()
+        return config.model_dump(mode="json", exclude_none=True)
 
     def _build_storage_config(self, s3_config: dict):
         storage_config = tempo_config.TraceStorage(
             # where to store the wal locally
-            wal=tempo_config.Wal(path=self.wal_path),
+            wal=tempo_config.Wal(path=self.wal_path),  # type: ignore
             pool=tempo_config.Pool(
                 # number of traces per index record
                 max_workers=400,
--- a/src/tempo.py
+++ b/src/tempo.py
@@ -4,17 +4,17 @@
 
 """Tempo workload configuration and client."""
 import logging
-import re
 import socket
 from subprocess import CalledProcessError, getoutput
-from typing import Dict, List, Optional, Sequence, Tuple
-from urllib.parse import urlparse
+from typing import Any, Dict, List, Optional, Sequence, Tuple
 
 from charms.tempo_k8s.v2.tracing import (
     ReceiverProtocol,
     receiver_protocol_to_transport_protocol,
 )
 from charms.traefik_route_k8s.v0.traefik_route import TraefikRouteRequirer
+
+import tempo_config
 
 logger = logging.getLogger(__name__)
 
@@ -31,10 +31,6 @@ class Tempo:
     tls_cert_path = "/etc/tempo/tls/server.crt"
     tls_key_path = "/etc/tempo/tls/server.key"
     tls_ca_path = "/usr/local/share/ca-certificates/ca.crt"
-
-    _tls_min_version = ""
-    # cfr https://grafana.com/docs/enterprise-traces/latest/configure/reference/#supported-contents-and-default-values
-    # "VersionTLS12"
 
     wal_path = "/etc/tempo/tempo_wal"
     log_path = "/var/log/tempo.log"
@@ -130,21 +126,21 @@ class Tempo:
         return f"{url}:{receiver_port}"
 
     def _build_server_config(self):
-        server_config = {
-            "http_listen_port": self.tempo_http_server_port,
+        server_config = tempo_config.Server(
+            http_listen_port=self.tempo_http_server_port,
             # we need to specify a grpc server port even if we're not using the grpc server,
             # otherwise it will default to 9595 and make promtail bork
-            "grpc_listen_port": self.tempo_grpc_server_port,
-        }
+            grpc_listen_port=self.tempo_grpc_server_port,
+        )
+
         if self.use_tls:
-            for cfg in ("http_tls_config", "grpc_tls_config"):
-                server_config[cfg] = {  # type: ignore
-                    "cert_file": str(self.tls_cert_path),
-                    "key_file": str(self.tls_key_path),
-                    "client_ca_file": str(self.tls_ca_path),
-                    "client_auth_type": "VerifyClientCertIfGiven",
-                }
-            server_config["tls_min_version"] = self._tls_min_version  # type: ignore
+            server_tls_config = tempo_config.TLS(
+                cert_file=str(self.tls_cert_path),
+                key_file=str(self.tls_key_path),
+                client_ca_file=str(self.tls_ca_path),
+            )
+            server_config.http_tls_config = server_tls_config
+            server_config.grpc_tls_config = server_tls_config
 
         return server_config
 
@@ -153,51 +149,21 @@ class Tempo:
         receivers: Sequence[ReceiverProtocol],
         s3_config: dict,
         peers: Optional[List[str]] = None,
-    ) -> dict:
+    ) -> Dict[str, Any]:
         """Generate the Tempo configuration.
 
         Only activate the provided receivers.
         """
-        config = {
-            "auth_enabled": False,
-            "server": self._build_server_config(),
-            # more configuration information can be found at
-            # https://github.com/open-telemetry/opentelemetry-collector/tree/overlord/receiver
-            "distributor": {"receivers": self._build_receivers_config(receivers)},
-            # the length of time after a trace has not received spans to consider it complete and flush it
-            # cut the head block when it hits this number of traces or ...
-            #   this much time passes
-            "ingester": {
-                "trace_idle_period": "10s",
-                "max_block_bytes": 100,
-                "max_block_duration": "30m",
-            },
-            "memberlist": {
-                "abort_if_cluster_join_fails": False,
-                "bind_port": self.memberlist_port,
-                "join_members": (
-                    [f"{peer}:{self.memberlist_port}" for peer in peers] if peers else []
-                ),
-            },
-            "compactor": {
-                "compaction": {
-                    # blocks in this time window will be compacted together
-                    "compaction_window": "1h",
-                    # maximum size of compacted blocks
-                    "max_compaction_objects": 1000000,
-                    # total trace retention
-                    "block_retention": "720h",
-                    "compacted_block_retention": "1h",
-                    "v2_out_buffer_bytes": 5242880,
-                }
-            },
-            # TODO this won't work for distributed coordinator where query frontend will be on a different unit
-            "querier": {
-                "frontend_worker": {"frontend_address": f"localhost:{self.tempo_grpc_server_port}"}
-            },
-            # see https://grafana.com/docs/tempo/latest/configuration/#storage
-            "storage": self._build_storage_config(s3_config),
-        }
+        config = tempo_config.Tempo(
+            auth_enabled=False,
+            server=self._build_server_config(),
+            distributor=self._build_distributor_config(receivers),
+            ingester=self._build_ingester_config(),
+            memberlist=self._build_memberlist_config(peers),
+            compactor=self._build_compactor_config(),
+            querier=self._build_querier_config(),
+            storage=self._build_storage_config(s3_config),
+        )
 
         if self.use_tls:
             # cfr:
@@ -210,42 +176,38 @@ class Tempo:
                 # try with fqdn?
                 "tls_server_name": self._external_hostname,
             }
-            config["ingester_client"] = {"grpc_client_config": tls_config}
-            config["metrics_generator_client"] = {"grpc_client_config": tls_config}
+            config.ingester_client = tempo_config.Client(
+                grpc_client_config=tempo_config.ClientTLS(**tls_config)
+            )
+            config.metrics_generator_client = tempo_config.Client(
+                grpc_client_config=tempo_config.ClientTLS(**tls_config)
+            )
+            config.querier.frontend_worker.grpc_client_config = tempo_config.ClientTLS(
+                **tls_config
+            )
+            config.memberlist = config.memberlist.model_copy(update=tls_config)
 
-            config["querier"]["frontend_worker"].update({"grpc_client_config": tls_config})
-
-            # this is not an error.
-            config["memberlist"].update(tls_config)
-
-        return config
+        return config.model_dump()
 
     def _build_storage_config(self, s3_config: dict):
-        storage_config = {
-            "wal": {
-                # where to store the wal locally
-                "path": self.wal_path
-            },
-            "pool": {
+        storage_config = tempo_config.TraceStorage(
+            # where to store the wal locally
+            wal=tempo_config.Wal(path=self.wal_path),
+            pool=tempo_config.Pool(
                 # number of traces per index record
-                "max_workers": 400,
-                "queue_depth": 20000,
-            },
-            "backend": "s3",
-            "s3": {
-                "bucket": s3_config["bucket"],
-                "access_key": s3_config["access-key"],
-                # remove scheme to avoid "Endpoint url cannot have fully qualified paths." on Tempo startup
-                "endpoint": re.sub(
-                    rf"^{urlparse(s3_config['endpoint']).scheme}://",
-                    "",
-                    s3_config["endpoint"],
-                ),
-                "secret_key": s3_config["secret-key"],
-                "insecure": (False if s3_config["endpoint"].startswith("https://") else True),
-            },
-        }
-        return {"trace": storage_config}
+                max_workers=400,
+                queue_depth=20000,
+            ),
+            backend="s3",
+            s3=tempo_config.S3(
+                bucket=s3_config["bucket"],
+                access_key=s3_config["access-key"],
+                endpoint=s3_config["endpoint"],
+                secret_key=s3_config["secret-key"],
+            ),
+            block=tempo_config.Block(version="v2"),
+        )
+        return tempo_config.Storage(trace=storage_config)
 
     def is_ready(self):
         """Whether the tempo built-in readiness check reports 'ready'."""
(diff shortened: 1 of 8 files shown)


src/charm.py
```diff
@@ -466,6 +466,9 @@ def _update_tempo_cluster(self):
             logger.error("skipped tempo cluster update: inconsistent state")
             return
 
+        if not self.unit.is_leader():
+            return
```
sed-i, line 470: Since you have a leader guard, do we need to make sure that the code below is run on `leader-elected`?

src/tempo.py
```diff
@@ -132,21 +151,21 @@ def get_receiver_url(self, protocol: ReceiverProtocol, ingress: TraefikRouteRequ
...
+
         if self.use_tls:
-            for cfg in ("http_tls_config", "grpc_tls_config"):
-                server_config[cfg] = {  # type: ignore
-                    "cert_file": str(self.tls_cert_path),
-                    "key_file": str(self.tls_key_path),
-                    "client_ca_file": str(self.tls_ca_path),
-                    "client_auth_type": "VerifyClientCertIfGiven",
-                }
-            server_config["tls_min_version"] = self._tls_min_version  # type: ignore
```
sed-i, line 149: Should we also remove the `_tls_min_version = ""` class attr?

src/tempo_cluster.py
```diff
@@ -161,10 +163,10 @@ def dump(self, databag: Optional[MutableMapping] = None, clear: bool = True):
                 databag = {}
 
             if self._NEST_UNDER:
-                databag[self._NEST_UNDER] = self.json(by_alias=True, exclude_defaults=True)
+                databag[self._NEST_UNDER] = self.json(by_alias=True, exclude_none=True)
```
sed-i, line 166: Do we still need `_NEST_UNDER`?
Iirc, it was introduced into traefik because at the time the old pydantic had some limitations.

src/tempo_config.py
sed-i, line 1: It's great that the config is ops independent. Let's try to keep it that way.

Re pydantic model itself (and how strict it is in loading rel data): how would an upgrade sequence look like when the new version has a schema change? Could we get tracebacks for validation error between coordinator and worker?

tests/integration/test_ingressed_tls.py
sed-i, line 1: Should we use a pytest "skip" call instead of commenting out the entire file?
Not sure which approach would be more noticeable down the road.

src/charm.py
michaeldmitry: Yes. It's already run on `leader-elected`

tests/integration/test_ingressed_tls.py
michaeldmitry: I think in terms of "noticeability", commenting the file is more visible. Anyways, its not going to hang like that for long. Should be uncommented, then fixed once we have a fully working cluster.

src/tempo_config.py
michaeldmitry: After discussion with @sed-i 
relation data for tempo config is sent as a dictionary, only the construction of the config in coordinator is done through pydantic models. And this dictionary is used by the dumb workers as is to run the workload.
So, in the event of changing the schema of the tempo configuration with a breaking change, workers would fail with an error in running the workload (or something similar). 
A suggestion was to try to validate the received dict on the model in workload side and if it fails, we would put the worker in Blocking state instead of crashing with an error.

@PietroPasotti  @mmkay what are your thoughts on this matter?

michaeldmitry: I think we don't even need to validate against the pydantic model.
https://github.com/grafana/tempo/issues/2131
Tempo itself can help us verify if the passed config file is correct or not.

mmkay (approved):

src/tempo_cluster.py
```diff
@@ -180,18 +174,12 @@ class DatabagModel(BaseModel):
...
-            # Custom config key: whether to nest the whole datastructure (as json)
-            # under a field or spread it out at the toplevel.
-            _NEST_UNDER=None,
         )  # type: ignore
         """Pydantic config."""
 
         @classmethod
         def load(cls, databag: MutableMapping):
             """Load this model from a Juju databag."""
-            nest_under = cls.model_config.get("_NEST_UNDER")
```
PietroPasotti, line 192: uhm, why did you remove the nest_under logic from the whole object?

src/tempo.py
```diff
@@ -132,21 +147,21 @@ def get_receiver_url(self, protocol: ReceiverProtocol, ingress: TraefikRouteRequ
         return f"{url}:{receiver_port}"
 
     def _build_server_config(self):
-        server_config = {
-            "http_listen_port": self.tempo_http_server_port,
+        server_config = Server(
```
PietroPasotti, line 150: all other types have a Config suffix, I think this should be ServerConfig as well.

src/tempo.py
```diff
@@ -212,42 +197,34 @@ def generate_config(
...
+            backend="s3",
+            s3=S3Config(
+                bucket=s3_config["bucket"],
+                access_key=s3_config["access-key"],
+                endpoint=s3_config["endpoint"],
+                secret_key=s3_config["secret-key"],
+            ),
+            block=BlockConfig(version="v2"),
+        )
+        return Storage(trace=storage_config)
```
PietroPasotti, line 227:
```suggestion
        return StorageConfig(trace=storage_config)
```

src/tempo.py
```diff
@@ -265,7 +242,51 @@ def is_ready(self):
             return False
         return out == "ready"
 
-    def _build_receivers_config(self, receivers: Sequence[ReceiverProtocol]):  # noqa: C901
+    def _build_querier_config(self):
+        """Build querier config"""
+        # TODO this won't work for distributed coordinator where query frontend will be on a different unit
+        return Querier(
```
PietroPasotti, line 248:
```suggestion
        return QuerierConfig(
```

src/tempo.py
```diff
@@ -265,7 +242,51 @@ def is_ready(self):
...
+            frontend_worker=FrontendWorker(
```
PietroPasotti, line 249:
```suggestion
            frontend_worker=FrontendWorkerConfig(
```

or, if you don't like the idea of having 100 objects all called SomethingConfig, I'd be fine with doing:
```
from xx import tempo_config
tempo_config.Querier
tempo_config.Server
```
and so on. Then the prefix makes it clear we're talking about the configuration of a server, and not the server itself.

src/tempo_cluster.py
```diff
@@ -22,6 +22,8 @@
 from ops import EventSource, Object, ObjectEvents
 from pydantic import BaseModel, ConfigDict
 
+from tempo_config import TempoBaseConfig
```
PietroPasotti, line 25: I'd rename this to `TempoConfig`. The name 'base' usually indicates you're meant to subclass it or specialize it somehow.

tests/integration/test_ingressed_tls.py
```diff
@@ -1,180 +1,177 @@
...
-async def test_remove_relation(ops_test: OpsTest):
-    await ops_test.juju(
-        "remove-relation", APP_NAME + ":certificates", SSC_APP_NAME + ":certificates"
-    )
-    await asyncio.gather(
-        ops_test.model.wait_for_idle(
-            apps=[APP_NAME], status="active", raise_on_blocked=True, timeout=1000
-        ),
-    )
+# import asyncio
```
PietroPasotti, line 1: why are we commenting out a whole module? 
If it's no longer relevant, let's remove it. 
If there's a really good reason to keep it around, write that in a FIXME/TODO.

tests/integration/test_charm.py
```diff
@@ -33,9 +33,6 @@ async def test_build_and_deploy(ops_test: OpsTest):
           {mc.name}:
             charm: {charm}
             trust: true
-            resources:
```
PietroPasotti, line 36: were these changes necessary?

tests/integration/test_integration.py
```diff
@@ -1,168 +1,165 @@
...
-        ops_test.model.wait_for_idle(
-            apps=[APP_NAME], status="active", raise_on_blocked=True, timeout=1000
-        ),
-        # for tester, depending on the result of race with tempo it's either waiting or active
-        ops_test.model.wait_for_idle(apps=[TESTER_APP_NAME], raise_on_blocked=True, timeout=1000),
-        ops_test.model.wait_for_idle(
-            apps=[TESTER_GRPC_APP_NAME], raise_on_blocked=True, timeout=1000
-        ),
-    )
+# import asyncio
```
PietroPasotti, line 1: same

tests/integration/test_scaling_monolithic.py
```diff
@@ -12,9 +13,10 @@
 from pytest_operator.plugin import OpsTest
 
 METADATA = yaml.safe_load(Path("./charmcraft.yaml").read_text())
-APP_NAME = "tempo"
+APP_NAME = "tempo-coordinator"
```
PietroPasotti, line 16: mimir's coordinator is also deployed as 'mimir' in the bundle. I'd not change this.

tests/integration/test_scaling_monolithic.py
```diff
@@ -65,19 +67,35 @@ def present_facade(
...
-    with tempfile.NamedTemporaryFile() as f:
+    with tempfile.NamedTemporaryFile(dir=os.getcwd()) as f:
         fpath = Path(f.name)
         fpath.write_text(yaml.safe_dump(data))
 
         _model = f" --model {model}" if model else ""
-        run(shlex.split(f"juju run {app}/0{_model} --params {fpath.absolute()}"))
+
+        run(shlex.split(f"juju run {app}/0{_model} update --params {fpath.absolute()}"))
+        # facade charm edge rev9 copies data into 'mocks/provide' not 'mocks/require'
```
PietroPasotti, line 77: this should be fixed now.

tests/integration/test_scaling_monolithic.py
```diff
@@ -23,13 +25,13 @@
...
-    await ops_test.model.deploy(tempo_charm, resources=resources, application_name=APP_NAME),
+    await ops_test.model.deploy(tempo_charm, application_name=APP_NAME)
 
     await ops_test.model.wait_for_idle(
         apps=[APP_NAME],
-        # tempo might be in waiting as it waits for tempo workload ready
-        raise_on_blocked=True,
+        # coordinator will be blocked on s3 and workers integration
+        status="blocked",
+        raise_on_blocked=False,
```
PietroPasotti, line 34: I _think_ if you expect status="blocked" it should also implicitly not raise on blocked.

tests/integration/test_scaling_monolithic.py
```diff
@@ -65,19 +67,35 @@ def present_facade(
...
 
         _model = f" --model {model}" if model else ""
-        run(shlex.split(f"juju run {app}/0{_model} --params {fpath.absolute()}"))
+
+        run(shlex.split(f"juju run {app}/0{_model} update --params {fpath.absolute()}"))
+        # facade charm edge rev9 copies data into 'mocks/provide' not 'mocks/require'
+        # workaround to mv the copied file to the correct path inside 'require' directory
+        # until charm-relation-interfaces/pull/152 is merged.
+        if role == "require":
+            run(
```
PietroPasotti, line 81: oh man so sorry you had to do this horrible thing :D

PietroPasotti (changes requested):
impressive work! 
I have a few suggestions about the API (class names, etc...) that are non-blocking
as for the commented-out modules, let's add an explanation of what's going on or let's remove them.

src/tempo.py
```diff
@@ -212,42 +197,34 @@ def generate_config(
...
-                "secret_key": s3_config["secret-key"],
-                "insecure": (False if s3_config["endpoint"].startswith("https://") else True),
-            },
-        }
-        return {"trace": storage_config}
+                max_workers=400,
+                queue_depth=20000,
+            ),
+            backend="s3",
+            s3=S3Config(
```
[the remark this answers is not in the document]
michaeldmitry: I don't think it would make much of a difference if you're referring to the same level of validation we currently do (keys exist with correct value types). 
once we get the `credentials_changed` event, we generate the config (`_update_tempo_cluster`). 
Also, I think it'd be cleaner to keep construction of config with the validation it offers in one place.

src/tempo_cluster.py
michaeldmitry: @sed-i raised a comment if we still need that since it might have been introduced due to old pydantic limitation.
Seems that with the pydantic version that gets used in this charm, things work normally without this object and its logic.

tests/integration/test_charm.py
michaeldmitry: Yes. resources "key" doesn't exist in this current charmcraft. it breaks the test.

src/tempo_config.py
PietroPasotti: So my thought here is we should try to fail early and loudly. 
**Option A: validate once.**
The coordinator generates a config, the pydantic model validates that it's correct. We put it in relation data.
Worker picks it up, puts it in the workload, workload breaks because the config is not valid.
Now we have to figure out that the workload is broken and why. 

**Option B: validate twice.**
The coordinator generates a config, the pydantic model validates that it's correct. We put it in relation data.
Worker picks it up, attempts to validate it but it fails. Worker immediately sets blocked status and goes belly up, saying 'config invalid'.

Regardless of _how_ we do the validation in the worker (whether it's by the built-in [cli tool offered by tempo](https://github.com/grafana/tempo/issues/2131) or by a copy of the same pydantic model hosted in the controller, that's up to us) I think we should.

**Option C: validate thrice.**
The coordinator generates a config, the pydantic model validates that it's correct. We put it in relation data.
Worker picks it up, attempts to validate it with (the same) pydantic model and it succeeds. 
It attempts to validate it with the tempo cli tool and it fails. 
This means that the pydantic model and the tempo image are out of sync: our models encode a config that's not supported in the tempo version we're running. 

Option C feels a bit like overkill, we can guard against this scenario by integration-testing the worker and verify that the config generated by the pydantic models always passes the built-in config validator. So I'd say let's go for B.
As to how we validate, I think Pydantic models offers us the most flexibility, because in case we want to be backwards-compatible or support consecutive tempo versions, the worker could implement logic like so:
```
if ConfigModelV1.validate(config_from_reldata):
    unit.status = ActiveStatus("Please upgrade the tempo coordinator to v2 (rev > 1412)")
   conf = ConfigModelV2.from_v1(config_from_reldata) 
else:
    conf = ConfigModelV2(config_from_reldata)
```

and the upgrade path would be:
- upgrade the workers
- upgrade the coordinator

src/tempo.py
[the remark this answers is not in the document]
PietroPasotti: I do like the idea to keep the validation in one place

tests/integration/test_ingressed_tls.py
PietroPasotti: how was this issue resolved exactly?

src/tempo_config.py
```diff
@@ -24,143 +22,184 @@ class InvalidConfigurationError(Exception):
...
 
-    ingestion_rate: int = 0
-    ingestion_burst_size: int = 0
-    max_global_series_per_user: int = 0
-    ruler_max_rules_per_rule_group: int = 0
-    ruler_max_rule_groups_per_tenant: int = 0
+    abort_if_cluster_join_fails: bool
+    bind_port: int
+    join_members: List[str]
+    tls_enabled: Optional[bool] = False
```
PietroPasotti, line 44: is this truly optional? as in, can it be None? or is it really bool (given that the default is False...)

src/tempo_config.py
```diff
@@ -24,143 +22,184 @@ class InvalidConfigurationError(Exception):
...
-    """Alertmanager schema."""
+class TLS(BaseModel):
+    """TLS configuration schema."""
 
-    data_dir: str = "/data/alertmanager"
-    external_url: Optional[str]
+    cert_file: str
+    key_file: str
+    client_ca_file: str
+    client_auth_type: Optional[str] = "VerifyClientCertIfGiven"
```
PietroPasotti, line 102: this feels like it should be an enum

src/tempo_config.py
```diff
@@ -24,143 +22,184 @@ class InvalidConfigurationError(Exception):
...
+    """Pool schema."""
+
+    max_workers: int
+    queue_depth: int
+
+
+class Wal(BaseModel):
+    """Wal schema."""
+
+    path: str
```
PietroPasotti, line 141: I think pydantic takes `pathlib.Path` as well? perhaps that gives better validation?

src/tempo_config.py
```diff
@@ -24,143 +22,184 @@ class InvalidConfigurationError(Exception):
...
+
+    bucket: str
+    access_key: str
     endpoint: str
-    secret_access_key: str = Field(alias="secret-key")
-    bucket_name: str = Field(alias="bucket")
-    region: str = ""
-    insecure: str = "false"
+    secret_key: str
+    insecure: Optional[bool] = False
```
PietroPasotti, line 151: just bool?

PietroPasotti (commented):
a few possible typing issues with the models

tests/integration/test_ingressed_tls.py
michaeldmitry: Added a TODO

src/tempo_config.py
michaeldmitry: its semantically misleading. Will make it a bool

src/tempo_config.py
```diff
@@ -3,164 +3,217 @@
...
 S3_RELATION_NAME = "s3"
 BUCKET_NAME = "tempo"
 
 logger = logging.getLogger(__name__)
 
 
-class InvalidConfigurationError(Exception):
-    """Invalid configuration."""
+class ClientAuthTypeEnum(str, enum.Enum):
+    """client auth types"""
```
PietroPasotti, line 22:
```suggestion
    """Client auth types."""
```

src/tempo_config.py
```diff
@@ -3,164 +3,217 @@
...
 
-    pass
+    VERIFY_CLIENT_CERT_IF_GIVEN = "VerifyClientCertIfGiven"
```
PietroPasotti, line 24: please add the other possible values:
https://pkg.go.dev/crypto/tls#ClientAuthType

PietroPasotti (approved):
once the last couple of issues with the enum are fixed, I'm fine with merging :) 
Thanks!

src/tempo_config.py
PietroPasotti: (and maybe add this link in a comment to document where we're grabbing these values from)

PR #14, opened 2024-07-17
Use coordinator + worker libs

michaeldmitry:
## Context 
Implement generic coordinator + worker.

## Testing Consideration
Run integration tests
```
tox -e integration
```


tests/integration/helpers.py
```diff
@@ -187,3 +202,108 @@ async def run_command(model_name: str, app_name: str, unit_num: int, command: li
...
+
+
+async def deploy_cluster(ops_test: OpsTest):
+    # await ops_test.model.deploy(FACADE, channel="edge")
+    # await ops_test.model.wait_for_idle(
+    #     apps=[FACADE], raise_on_blocked=True, status="active", timeout=2000
+    # )
+
+    # TODO: deploy from latest edge
+    tempo_worker_charm = "/home/michael/Work/tempo-worker-k8s-operator/charm"
```
lucabello, line 288: Remember to not merge this :)

tests/integration/helpers.py
```diff
@@ -187,3 +206,149 @@ async def run_command(model_name: str, app_name: str, unit_num: int, command: li
...
+        f"{minio_addr}:9000",
+        access_key="accesskey",
+        secret_key="secretkey",
+        secure=False,
+    )
+
+    # create tempo bucket
+    found = mc_client.bucket_exists(BUCKET_NAME)
+    if not found:
+        mc_client.make_bucket(BUCKET_NAME)
```
mmkay, line 267: Nice find :)

tests/integration/helpers.py
```diff
@@ -187,3 +206,149 @@ async def run_command(model_name: str, app_name: str, unit_num: int, command: li
...
+    # await ops_test.model.deploy(FACADE, channel="edge")
+    # await ops_test.model.wait_for_idle(
+    #     apps=[FACADE], raise_on_blocked=True, status="active", timeout=2000
+    # )
+
+    # TODO: deploy from latest edge
+    tempo_worker_charm = "/home/michael/Work/tempo-worker-k8s-operator/charm"
+
+    resources = {
+        "tempo-image": "docker.io/ubuntu/tempo:2-22.04",
```
mmkay, line 295: Could we use the image from the metadata?

tests/integration/test_ingressed_tls.py
```diff
@@ -1,138 +1,97 @@
...
+    if "Serving at" not in status_msg:
+        assert False, f"Ingressed hostname is not present in {TRAEFIK_APP_NAME} status message."
+    return status_msg.replace("Serving at", "").strip()
+
+
+@pytest.mark.setup
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest):
+    # tempo_charm = await ops_test.build_charm(".")
+    tempo_charm = "/home/michael/Work/tempo-coordinator-k8s-operator/charm"
```
mmkay, line 38: Same as Luca's comment above :)

tests/integration/test_integration.py
```diff
@@ -1,166 +1,168 @@
...
+
+
+@pytest.mark.setup
+@pytest.mark.abort_on_fail
+async def test_build_and_deploy(ops_test: OpsTest):
+    # Given a fresh build of the charm
+    # When deploying it together with testers
+    # Then applications should eventually be created
+    # tempo_charm = await ops_test.build_charm(".")
+    tempo_charm = "/home/michael/Work/tempo-coordinator-k8s-operator/charm"
```
mmkay, line 30: ...and here :)

tests/integration/test_scaling_monolithic.py
```diff
@@ -1,30 +1,26 @@
...
 
 logger = logging.getLogger(__name__)
 
 
 @pytest.mark.setup
 @pytest.mark.abort_on_fail
 async def test_deploy_tempo(ops_test: OpsTest):
-    tempo_charm = await ops_test.build_charm(".")
+    # tempo_charm = await ops_test.build_charm(".")
+    tempo_charm = "/home/michael/Work/tempo-coordinator-k8s-operator/charm"
```
mmkay, line 22: ... and here :)

mmkay (commented):
A round of comments from me, mostly about test code.

src/charm.py
```diff
@@ -195,76 +141,12 @@ def _external_url(self) -> str:
...
-        """Return True if tls is enabled and the necessary certs are found."""
-        return (
-            self.cert_handler.enabled
-            and (self.cert_handler.server_cert is not None)
-            and (self.cert_handler.private_key is not None)
-            and (self.cert_handler.ca_cert is not None)
-        )
+        """Returns workload's FQDN."""
+        scheme = "http"
+        if hasattr(self, "coordinator") and self.coordinator.nginx.are_certificates_on_disk:
```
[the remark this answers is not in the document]
michaeldmitry: I added the `hasattr(...)` to places where `@trace_charm` calls them since they might be called before the `self.coordinator` gets instantiated in `__init__`. I might be wrong. WDYT?

src/charm.py
```diff
@@ -459,95 +268,37 @@ def _requested_receivers(self) -> Tuple[ReceiverProtocol, ...]:
...
         else:  # tls unavailable: delete local cert
-            server_cert.unlink(missing_ok=True)
+            server_ca_cert.unlink(missing_ok=True)
 
     def tempo_otlp_http_endpoint(self) -> Optional[str]:
         """Endpoint at which the charm tracing information will be forwarded."""
         # the charm container and the tempo workload container have apparently the same
         # IP, so we can talk to tempo at localhost.
-        if self.tempo.is_ready:
+        if hasattr(self, "coordinator") and self.coordinator.tracing.is_ready():
```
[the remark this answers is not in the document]
michaeldmitry: if the assumption in the above comment is correct, then self.coordinator would raise an `AttributeError` exception if it gets called before self.coordinator gets initialised.

tests/integration/helpers.py
michaeldmitry: No. Since the metadata will be for the Tempo coordinator charm not the worker. Anyways, this line will be removed since we'll deploy worker from edge.

src/tempo.py
```diff
@@ -201,39 +141,22 @@ def _build_storage_config(self, s3_config: dict):
...
-        except (CalledProcessError, IndexError):
-            return False
-        return out == "ready"
-
     def _build_querier_config(self, roles_addresses: Dict[str, Set[str]]):
         """Build querier config"""
-        addr = "localhost"
-        if TempoRole.query_frontend in roles_addresses.keys():
-            addr = roles_addresses[TempoRole.query_frontend].pop()
+        # if distributor and query-frontend have the same address, then the mode of operation is 'all'.
```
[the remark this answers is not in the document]
michaeldmitry: Well, in that case it will fetch the `query_frontend` address from the address set of `query_frontend` and not use `localhost`. Since the set of addrs for distributors will be different than the set of addrs for query frontends. 
This can only be one given address not a list, as Tempo supports.

src/tempo_config.py
```diff
@@ -2,22 +2,96 @@
...
+}
+"""Tempo component meta-role names."""
+
+MINIMAL_DEPLOYMENT = {
+    TempoRole.querier: 1,
+    TempoRole.query_frontend: 1,
+    TempoRole.ingester: 1,
+    TempoRole.distributor: 1,
+    TempoRole.compactor: 1,
+    TempoRole.metrics_generator: 1,
```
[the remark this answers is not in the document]
michaeldmitry: We already have that https://github.com/canonical/tempo-coordinator-k8s-operator/blob/main/src/coordinator.py#L15 in the minimal deployment. 
However, from [Grafana Tempo docs](https://grafana.com/docs/tempo/latest/operations/architecture/), it seems to be an "Optional" component. 
Will remove it and test.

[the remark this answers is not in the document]
michaeldmitry: even in helm charts, its not enabled by default. https://github.com/grafana/helm-charts/blob/main/charts/tempo-distributed/values.yaml.

scripts/tracegen.py
```diff
@@ -36,12 +36,12 @@ def emit_trace(
...
 
-    return _export_trace(span_exporter, log_trace_to_console=log_trace_to_console, nonce=nonce)
+    return _export_trace(span_exporter, log_trace_to_console=log_trace_to_console, nonce=nonce, protocol = protocol)
 
 
-def _export_trace(span_exporter, log_trace_to_console: bool = False, nonce: Any = None):
+def _export_trace(span_exporter, log_trace_to_console: bool = False, nonce: Any = None, protocol: Literal["grpc", "http"] = "http"):
     resource = Resource.create(attributes={
-        "service.name": "tracegen",
+        "service.name": f"tracegen{protocol}",
```
PietroPasotti, line 44: good idea!

src/charm.py
```diff
@@ -51,29 +42,28 @@ def __init__(self, *args):
...
+        # set the open ports for this unit
+        self.unit.set_ports(*self.tempo.all_ports.values())
+        self.coordinator = Coordinator(
+            charm=self,
+            roles_config=TempoRolesConfig(),
+            s3_bucket_name=Tempo.s3_bucket_name,
+            external_url=self._external_url,
+            worker_metrics_port=8080,
+            endpoints={
+                "certificates": "certificates",
```
PietroPasotti, line 55: IIRC we only (should) need to override those that are different from the default. That is only `metrics`, `cluster`, and `tracing` in our case.
Right @lucabello ?

src/charm.py
```diff
@@ -82,91 +72,47 @@ def __init__(self, *args):
...
-                "the situation is resolved by the cloud admin, to avoid data loss."
-            )
-            return  # refuse to handle any other event as we can't possibly know what to do.
+        # refuse to handle any other event as we can't possibly know what to do.
+        if (
+            not self.coordinator.cluster.has_workers
+            or not self.coordinator.is_coherent
+            or not self.coordinator.s3_ready
+        ):
+            return
```
PietroPasotti, line 92: I assume we're not logging anything here because the coordinator will do that for us under the same circumstances? (let's note this in a comment)

if so, we should consider exposing this conditions from the coordinator. 

How about a `Coordinator.can_handle_events --> self.has_workers & self.is_coherent & self.s3_ready`?

@lucabello

src/charm.py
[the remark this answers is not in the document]
PietroPasotti: `trace_charm` always calls `__init__` first, so this should not be necessary.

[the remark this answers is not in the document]
PietroPasotti: (if you look at the previous code, we'd have had the same problem with `self.tls_available`)

src/charm.py
```diff
@@ -288,40 +170,26 @@ def _on_tracing_broken(self, _):
...
-                server_cert=self.cert_handler.server_cert,  # type: ignore
-                ca_cert=self.cert_handler.ca_cert,  # type: ignore
-                private_key=self.cert_handler.private_key,  # type: ignore
-            )
-        else:
-            logger.debug("disabling TLS")
-            self.nginx.delete_certificates()
+    def _on_cert_handler_changed(self, e: ops.RelationChangedEvent):
+
+        self.coordinator._on_cert_handler_changed(e)
```
PietroPasotti, line 175: I would expect the coordinator to do that on its own while handling the event? 
This handler will be called after the coordinator's one runs, so we should be covered.

src/charm.py
```diff
@@ -366,52 +211,16 @@ def _on_ingress_revoked(self, _event):
...
-        self.nginx.configure_pebble_layer()
-        self.nginx_prometheus_exporter.configure_pebble_layer()
-        # configure ingress
-        self._configure_ingress()
-        # update cluster relations
-        self._update_tempo_cluster()
-        # update tracing relations
-        self._update_tracing_relations()
+    def _on_collect_unit_status(self, e: ops.CollectStatusEvent):
+        self.coordinator._on_collect_unit_status(e)
```
PietroPasotti, line 219: again, doesn't the coordinator also listen to the same event and handle this already?

src/charm.py
```diff
@@ -429,7 +238,7 @@ def _configure_ingress(self) -> None:
                 self._update_tracing_relations()
 
         # notify the cluster
-        self._update_tempo_cluster()
+        self.coordinator.update_cluster()
```
PietroPasotti, line 241: @lucabello 
as you see here we need to call update_cluster like it were a common exit hook after a bunch of different events.
Would it make sense to add to Coordinator something like a `cluster_refresh_events` list of events that the coordinator?

The only caveat to this approach would be that we'd need to call update_cluster AFTER the event processing has executed, on commit, because otherwise the framework will queue that before the subclass' own handlers.

src/charm.py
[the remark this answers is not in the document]
PietroPasotti: like above, I think it's fine to do `self.coordinator`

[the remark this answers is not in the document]
PietroPasotti: (if it's not you would definitely catch this in unittests)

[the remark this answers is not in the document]
PietroPasotti: (assuming we have at least one charm-tracing-enabled unittest)

src/charm.py
```diff
@@ -590,6 +340,49 @@ def _ingress_config(self) -> dict:
             },
         }
 
+    def get_receiver_url(self, protocol: ReceiverProtocol):
```
PietroPasotti, line 343: didn't we have a very similar piece of code in the tracing library (`get_endpoint`)? Can you check if we can avoid duplicating that by lifting it out to a utility method we can invoke from here?

src/tempo.py
```diff
@@ -201,39 +141,22 @@ def _build_storage_config(self, s3_config: dict):
                 queue_depth=20000,
             ),
             backend="s3",
-            s3=tempo_config.S3(
-                bucket=s3_config["bucket"],
-                access_key=s3_config["access-key"],
-                endpoint=s3_config["endpoint"],
-                secret_key=s3_config["secret-key"],
-            ),
+            s3=tempo_config.S3(**s3_config),
```
PietroPasotti, line 144: beware of https://github.com/canonical/cos-lib/pull/36

src/tempo.py
```diff
@@ -201,39 +141,22 @@ def _build_storage_config(self, s3_config: dict):
...
-        addr = "localhost"
-        if TempoRole.query_frontend in roles_addresses.keys():
-            addr = roles_addresses[TempoRole.query_frontend].pop()
+        # if distributor and query-frontend have the same address, then the mode of operation is 'all'.
+        if roles_addresses.get(tempo_config.TempoRole.query_frontend) == roles_addresses.get(
+            tempo_config.TempoRole.distributor
+        ) or not roles_addresses.get(tempo_config.TempoRole.query_frontend):
+            addr = "localhost"
+        else:
+            addr = roles_addresses.get(tempo_config.TempoRole.query_frontend, set()).pop()
```
PietroPasotti, line 159: we already check for emptiness in the if clause, we could do a straight getitem here.

src/tempo.py
```diff
@@ -201,39 +141,22 @@ def _build_storage_config(self, s3_config: dict):
...
-            return False
-        return out == "ready"
-
     def _build_querier_config(self, roles_addresses: Dict[str, Set[str]]):
         """Build querier config"""
-        addr = "localhost"
-        if TempoRole.query_frontend in roles_addresses.keys():
-            addr = roles_addresses[TempoRole.query_frontend].pop()
+        # if distributor and query-frontend have the same address, then the mode of operation is 'all'.
+        if roles_addresses.get(tempo_config.TempoRole.query_frontend) == roles_addresses.get(
```
PietroPasotti, line 154: proposal:

```python
query_frontend_addresses = roles_addresses.get(tempo_config.TempoRole.query_frontend)
distributor_addresses = roles_addresses.get(tempo_config.TempoRole.distributor)

if not query_frontend_addresses or query_frontend_addresses  == distributor_addresses:
    addr = "localhost"
else: 
    addr = query_frontend_addresses.pop()
```

PietroPasotti (commented):
gave a review focusing on the charm code

src/charm.py
lucabello: This is not the case anymore; the double types `_EndpointMapping` and `_EndpointMappingOverride` was found confusing by the team, and we ended up removing it in favor of just `_EndpointMapping`, and explicitly specifying the endpoints for all the relations

src/charm.py
lucabello: I think it's a good idea! I have a PR open in cos-lib, I'll sneak this in!

src/charm.py
lucabello: No preference really, I'd say let's see what happens in all the three HA solutions, and we can refactor it afterwards, maybe?

Calling the function is a bit more "explicit", but we can do whichever

src/charm.py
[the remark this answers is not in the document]
michaeldmitry: I think the issue is with `self._external_url` that gets passed to `Coordinator(external_url=self._external_url)`. That flow will eventually lead to `_internal_url` that will try to access `self.coordinator` that is not yet initialised. 
I think this might be an even broader problem? that `_external_url` will be passed to `MetricsEndpointProvider` and since `self.coordinator` will not be there yet, in the case of TLS, `MetricsEndpointProvider` will always get an http instead of https endpoint.

src/charm.py
michaeldmitry: The one we have in the lib `get_endpoint` gets the receivers endpoints from the `TracingRequirerAppData` databag. This data gets published using the `get_receiver_url` in charm code. So, this function is actually the source.

src/tempo.py
michaeldmitry: After checking, the `coordinator::_s3_config` is usable for Mimir, but not Tempo. In Tempo, S3 config has `secret_key` and `access_key` not `secret_access_key` and `access_key_id`

src/charm.py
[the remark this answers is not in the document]
PietroPasotti: ah. Well yes, that is definitely a problem. 
why does coordinator need external_url as an argument, given it wraps ingress as well? @lucabello

src/charm.py
PietroPasotti: Aha. I see it's a typed dict now. Thanks for clarifying :)

src/charm.py
PietroPasotti: uh right, I may be confusing it from a point in time when this logic was living in the charm lib. We refactored it out of it at some point I guess. Thanks for checking :)

src/tempo.py
PietroPasotti: haha lovely. Let's make a note then of why we're needing this transform step.

src/charm.py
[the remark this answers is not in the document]
michaeldmitry: it doesn't?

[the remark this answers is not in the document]
michaeldmitry: I just realized I was writing on the wrong comment :) I'm referring to line 147
Shared coordinator doesn't wrap ingress. I think we can just remove the dependency on `self.coordinator.nginx.are_certificates_on_disk` and have a property in charm code that does that check for us.
```
    @property
    def are_certificates_on_disk(self) -> bool:
        """Return True if the certificates files are on disk."""
        nginx_container = self.unit.get_container("nginx")

        return (
            nginx_container.can_connect()
            and nginx_container.exists(CERT_PATH)
            and nginx_container.exists(KEY_PATH)
            and nginx_container.exists(CA_CERT_PATH)
        )

```
WDYT?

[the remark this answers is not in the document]
lucabello: The coordinator object does **not** wrap `IngressPerAppRequirer` specifically because apparently Tempo won't have that integration.

Correct me if I'm wrong, but shouldn't this same issue appear with anything that implements Ingress? All those charms have:

```python
@property
def external_url:
	return self.ingress.url or self.internal_url

def tempo_otlp_http_endpoint(self) -> Optional[str]:
	return self.external_url
```

Shouldn't this error for any charm related to ingress, considering "ingress" might not be yet initialized? This feels like something which is either a code ordering issue (which somehow doesn't happen in other places), or something that should be "fixed" in the tracing library itself if possible

Or maybe I'm just not clear on what's happening :)

PR #28, opened 2024-08-29
Enable `metrics-generator`

michaeldmitry:
This PR adds support for enabling the `metrics-generator` HA component.
- Add an integration over `prometheus_remote_write` to retrieve `remote-write` endpoints.
- Populate `metrics-generator` config with the endpoints.
- Set a coordinator unit status message to notify that `metrics-generator` is disabled and requires an integration. 

## Context
- `metrics-generator` requires a `remote write` endpoint(s) to push the generated metrics to. 
- https://github.com/canonical/cos-lib/pull/64
- [Worker](https://github.com/canonical/tempo-worker-k8s-operator/pull/25)

## Testing Instructions
Scenario tests


charmcraft.yaml
```diff
@@ -93,6 +93,10 @@ requires:
     description: |
       Certificate and key files for securing Tempo internal and external 
       communications with TLS.
+  send-remote-write:
+    interface: prometheus_remote_write
+    description: |
+      Prometheus-like remote write endpoints to push traces' metrics generated from `metrics-generator`.
```
PietroPasotti, line 99:
```suggestion
      Prometheus-like remote write endpoints to push traces' metrics generated by the `metrics-generator` component.
```

lib/charms/observability_libs/v0/kubernetes_compute_resources_patch.py
PietroPasotti, line 1: is this a leftover from another branch?

src/charm.py
```diff
@@ -44,9 +48,13 @@ def __init__(self, *args):
         super().__init__(*args)
 
         self.ingress = TraefikRouteRequirer(self, self.model.get_relation("ingress"), "ingress")  # type: ignore
+        self._remote_write = PrometheusRemoteWriteConsumer(self, alert_rules_path="")
```
PietroPasotti, line 51: can you add a comment explaining what the `alert_rules_path=""` is about?

src/charm.py
```diff
@@ -220,6 +235,20 @@ def _on_list_receivers_action(self, event: ops.ActionEvent):
...
+        """Event handler for the remote write changed event."""
+        # notify the cluster
+        self.coordinator.update_cluster()
+
+    def _on_collect_status(self, e: CollectStatusEvent):
+        # add Tempo coordinator-specific statuses
+        if not self.remote_write_endpoints():
+            e.add_status(
+                ops.ActiveStatus(
+                    "metrics-generator disabled. Add a relation over send-remote-write"
```
PietroPasotti, line 248: this is only relevant if we have a worker node with metrics-generator enabled.
IMHO we should be doing something like:
```
if 'metrics-generator' in self.coordinator.worker_roles and not self.remote_write_endpoints():
    ... set active
```

PietroPasotti (approved):
Couple of minor nits, but looks good!

PR #32, opened 2024-09-05
Update interface testers and tests for tempo-cluster and tracing

mmkay:


PietroPasotti (approved):

tests/interface/test_tempo_cluster.py
```diff
@@ -0,0 +1,11 @@
+# Copyright 2022 Canonical Ltd.
```
dimaqq, line 1: Odd copyright year for a new file... I mean even, but still odd 🙈

PR #38, opened 2024-09-10
Speed-up integration tests 

michaeldmitry:
Integration tests take a long time to run (~2hrs). There are areas of improvement, specifically:
- Combining several co-related tests into one to save the build, deploy, and integration time.
- Combining setting config options into one config set instead of setting each config on its own and waiting for active/idle state after each config set.


PietroPasotti:
another easy win is only pack once instead of once per module

PR #47, opened 2024-09-25
set blocked if tempo has tls relation but traefik hasn't

PietroPasotti:
without this, it's hard to see that tls is misconfigured.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -203,6 +203,23 @@ class TempoCoordinatorCharm(CharmBase):
                     "metrics-generator disabled. Add a relation over send-remote-write"
                 )
             )
+        if (
+            # we have tls
+            self.coordinator.tls_available
+            # we have tls
+            and self.ingress.is_ready()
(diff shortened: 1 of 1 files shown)


src/charm.py
```diff
@@ -203,6 +203,23 @@ def _on_collect_status(self, e: CollectStatusEvent):
                     "metrics-generator disabled. Add a relation over send-remote-write"
                 )
             )
+        if (
+            # we have tls
+            self.coordinator.tls_available
+            # we have tls
+            and self.ingress.is_ready()
+            and self.ingress.scheme == "http"
```
sed-i, line 211: That's what we called "reverse termination".
- Strictly speaking it's not a bug
- I'd put this check on traefik side (lb server is https but traefik itself isn't)

PietroPasotti: that makes sense

PietroPasotti: opened an issue in traefik to track, closing this one

PR #53, opened 2024-09-30
s3 tls support

PietroPasotti:
- added region to s3 data model in cluster relation
- removed validators since cos-lib does that already for us
- added s3_ca_path to data model

released on latest/edge/pc6test


src/tempo_config.py
```diff
@@ -255,25 +253,10 @@ class S3(BaseModel):
     bucket_name: str = Field(alias="bucket")
     access_key_id: str = Field(alias="access_key")
     endpoint: str
+    region: Optional[str] = None
```
mmkay, line 256: Idea: could we put a `default` here?

michaeldmitry (approved):
We don't need `git` for the build-packages, but other than that LGTM :)

src/tempo_config.py
PietroPasotti: Ideally I'd like to avoid having defaults in multiple places, and I'm also not sure all s3 providers will behave the same way if we pass different region names. assuming ceph likes the "default" value, it's hard to tell whether amazon will like it too. I think ultimately it's up to the user to configure the s3 integrator the right way depending on the backend

PR #57, opened 2024-10-09
Charm tracing cache

PietroPasotti:
adds a buffer to charm_tracing, so that charm traces caught while tracing is unavailable can be kept on a local db and flushed once a tracing relation becomes available.

This includes:
- setup phase events (start, install...) when the charm hasn't been exposed yet to tracing relation data
- any events occurring before the user chooses to integrate with a tracing provider
- any events occurring in transitional phases when tracing is misconfigured, e.g. when you're adding TLS and the charm still has to write a cert to disk

image

The buffer has two main mechanisms for preventing OOM situations if the tracing integration is never added:
- a 'max number of events to cache' (default 100)
- a 'max cache size in MB' (also default 100).

Whichever is hit first will trigger the buffer to start dropping events (first first, last last) until the limits are met again.


lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py
```diff
@@ -543,7 +749,38 @@ def wrap_close():
...
+                    else:
+                        # if the buffer was nonempty, we can attempt to flush it
+                        buffer_flush_successful = buffer.flush()
+                        if buffer_flush_successful:
+                            logger.debug("buffer flush OK")
+                        elif buffer_flush_successful is None:
+                            # TODO is this even possible?
+                            logger.debug("buffer flush OK; empty: nothing to flush")
+                        else:
+                            # this situation is pretty weird, I'm not even sure it can happen,
```
mmkay, line 778: on connectivity drops maybe? 

that's also makes me wonder: what if we already have a connection to the tracing backend, we manage to send some traces, but then the connection breaks / tracing backend goes down? Will buffer already be able to deal with that?

lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py
```diff
@@ -290,6 +304,178 @@ def _remove_stale_otel_sdk_packages():
...
+    We cannot store them as json because that is not well-supported by the sdk
+    (see https://github.com/open-telemetry/opentelemetry-python/issues/3364).
+    """
+
+    _SPANSEP = b"__CHARM_TRACING_BUFFER_SPAN_SEP__"
+
+    def __init__(self, db_file: Path, max_event_history_length: int, max_buffer_size_mb: int):
+        self._db_file = db_file
+        self._max_event_history_length = max_event_history_length
+        self._max_buffer_size_mb = max(max_buffer_size_mb, 10)
```
mmkay, line 330: how about a constant?

lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py
```diff
@@ -290,6 +304,178 @@ def _remove_stale_otel_sdk_packages():
...
+class _BufferedExporter(InMemorySpanExporter):
+    def __init__(self, buffer: _Buffer) -> None:
+        super().__init__()
+        self._buffer = buffer
+
+    def export(self, spans: typing.Sequence[ReadableSpan]) -> SpanExportResult:
+        self._buffer.save(spans)
+        return super().export(spans)
+
+    def force_flush(self, timeout_millis: int = 30000) -> bool:
```
mmkay, line 475: Should the timeout be a constant? it's also 30s, sounds like a lot

lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py
```diff
@@ -474,33 +656,57 @@ def wrap_init(self: CharmBase, framework: Framework, *args, **kwargs):
...
+            # we have to buffer because we're missing necessary backend configuration
+            logger.debug("buffering ON")
+            exporters.append(_BufferedExporter(buffer))
+
+        else:
+            # in principle, we have the right configuration to be pushing traces
+            otlp_exporter = _OTLPSpanExporter(
+                endpoint=tracing_endpoint,
+                certificate_file=str(Path(server_cert).absolute()) if server_cert else None,
+                timeout=1,  # give individual requests 1 second to succeed
```
mmkay, line 700: I'd probably try to refactor these magic numbers into constants

mmkay (changes requested):
This is already an amazing thing but I think there's a couple final touches that might be needed. Please have a look at the comments :)

lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py
```diff
@@ -290,6 +304,178 @@ def _remove_stale_otel_sdk_packages():
...
+        new = self._serialize(spans)
+
+        try:
+            # if the buffer exceeds the size limit, we start dropping old spans until it does
+
+            while len((new + self._SPANSEP.join(old))) > (self._max_buffer_size_mb * _MB_TO_B):
+                if not old:
+                    # if we've already dropped all spans and still we can't get under the
+                    # size limit, we can't save this span
+                    logger.error(
```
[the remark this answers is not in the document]
PietroPasotti: we adjusted this right now, and we boiled it down to two error scenarios:
- your charm has a tracing integration, but for some reason we can't push traces to the backend
- your charm has a tracing integration, and we can send traces to it, but we can't flush the buffered ones

we think both cases should be errors on the regular logger, because in the first case there could be a glitch in your network/configuration, in the second it's likely a bug in the charm tracing lib itself we'd want to know about.
all other logs are for the devlogger only

lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py
PietroPasotti: this is hardcoded in this way in the method I'm overriding, so giving it a different default would anger pyright, besides, the parent implementation literally is `return True` as this is in fact a dummy exporter.

PietroPasotti: but I'll add a note

lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py
```diff
@@ -290,6 +304,178 @@ def _remove_stale_otel_sdk_packages():
...
+        buffered_spans = self.load()
+        if not buffered_spans:
+            logger.debug("nothing to flush; buffer empty")
+            return None
+
+        errors = False
+        for span in buffered_spans:
+            try:
+                out = self.exporter._export(span)  # type: ignore
+                if out.status_code not in (200, 202):
```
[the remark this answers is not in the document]
PietroPasotti: not sure that's ever going to happen, but indeed, better safe than sorry

lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py
```diff
@@ -682,22 +741,25 @@ def wrap_init(self: CharmBase, framework: Framework, *args, **kwargs):
...
         exporters: List[SpanExporter] = []
         if buffer_only:
             # we have to buffer because we're missing necessary backend configuration
-            logger.debug("buffering ON")
+            logger.debug("buffering mode: ON")
             exporters.append(_BufferedExporter(buffer))
 
         else:
-            # in principle, we have the right configuration to be pushing traces
+            logger.debug("buffering mode: FALLBACK")
```
mmkay, line 755: `dev_logger` perhaps? Same for most of debug logs around.

mmkay (approved):
Yes, that looks great :) One minor comment, feel free to resolve if you don't agree.

PR #59, opened 2024-10-11
load-balance between scaled-up coordinator units

michaeldmitry:
## Issue
When the coordinator is scaled up, this doesn't reflect in the ingress configuration of `traefik_route`. So, requests to traefik will always be proxied to one unit of the coordinator although we might have multiple units in place.

## Solution
Listen on `peer_relation_created` hook to populate peer data with hostnames of each coordinator unit and when constructing the ingress configuration, we retrieve the servers urls from the peer data.

 ## Drive-bys
 - Remove any DNS reference in the worker config to the coordinator or traefik DNS names and depend solely on the workers' DNS names. The rationale behind that is that the worker config sent to a worker should indicate that this config is sufficient and we can run a worker process with no need for sth outside the worker to be ready. i.e: If we pass traefik's hostname as `tls_server_name` in the worker config, at any point in time, traefik can now be behind TLS and our worker will suddenly fail as it needs Traefik's CA..
 - Update `CertHandler` lib
 
 ## Testing Considerations
 1. deploy this bundle
 ```
bundle: kubernetes
applications: 
  ssc:
    charm: self-signed-certificates
    channel: latest/edge
    trust: true
  minio:
    charm: minio
    channel: latest/edge
    options:
      access-key: accesskey
      secret-key: mysoverysecretkey
    trust: true
  traefik:
    charm: traefik-k8s
    channel: latest/edge
    scale: 1
    constraints: arch=amd64
    trust: true
  tempo:
    charm: /home/michael/Work/tempo-coordinator-k8s-operator/tempo-coordinator-k8s_ubuntu-22.04-amd64.charm
    scale: 3
    # constraints: arch=amd64 tags=anti-pod.app.kubernetes.io/name=tempo,anti-pod.topology-key=kubernetes.io/hostname
    trust: true
    resources:
      nginx-image: ubuntu/nginx:1.24-24.04_beta
      nginx-prometheus-exporter-image: nginx/nginx-prometheus-exporter:1.1.0
  s3-integrator:
    charm: s3-integrator
    channel: latest/edge
    scale: 1
    constraints: arch=amd64
  tempo-compactor:
    charm: tempo-worker-k8s
    channel: latest/edge
    resources:
      tempo-image: 3
    scale: 3
    options:
      role-all: false
      role-compactor: true
    # constraints: arch=amd64 tags=anti-pod.app.kubernetes.io/name=tempo-compactor,anti-pod.topology-key=kubernetes.io/hostname
    storage:
      data: kubernetes,1,1024M
    trust: true
  tempo-distributor:
    charm: tempo-worker-k8s
    channel: latest/edge
    resources:
      tempo-image: 3
    scale: 3
    options:
      role-all: false
      role-distributor: true
    # constraints: arch=amd64 tags=anti-pod.app.kubernetes.io/name=tempo-distributor,anti-pod.topology-key=kubernetes.io/hostname
    storage:
      data: kubernetes,1,1024M
    trust: true
  tempo-ingester:
    charm: tempo-worker-k8s
    channel: latest/edge
    resources:
      tempo-image: 3
    scale: 3
    options:
      role-all: false
      role-ingester: true
    # constraints: arch=amd64 tags=anti-pod.app.kubernetes.io/name=tempo-ingester,anti-pod.topology-key=kubernetes.io/hostname
    storage:
      data: kubernetes,1,1024M
    trust: true
  tempo-metrics-generator:
    charm: tempo-worker-k8s
    channel: latest/edge
    resources:
      tempo-image: 3
    scale: 3
    options:
      role-all: false
      role-metrics-generator: true
    # constraints: arch=amd64 tags=anti-pod.app.kubernetes.io/name=tempo-metrics-generator,anti-pod.topology-key=kubernetes.io/hostname
    storage:
      data: kubernetes,1,1024M
    trust: true
  tempo-querier:
    charm: tempo-worker-k8s
    channel: latest/edge
    resources:
      tempo-image: 3
    scale: 3
    options:
      role-all: false
      role-querier: true
    # constraints: arch=amd64 tags=anti-pod.app.kubernetes.io/name=tempo-querier,anti-pod.topology-key=kubernetes.io/hostname
    storage:
      data: kubernetes,1,1024M
    trust: true
  tempo-query-frontend:
    charm: tempo-worker-k8s
    channel: latest/edge
    resources:
      tempo-image: 3
    scale: 3
    options:
      role-all: false
      role-query-frontend: true
    # constraints: arch=amd64 tags=anti-pod.app.kubernetes.io/name=tempo-query-frontend,anti-pod.topology-key=kubernetes.io/hostname
    storage:
      data: kubernetes,1,1024M
    trust: true
```
 2. run `juju run s3-integrator/leader sync-s3-credentials access-key=accesskey secret-key=mysoverysecretkey`
 3. create a `tempo` bucket
 4. run `juju config s3-integrator endpoint=minio-0.minio-endpoints.test.svc.cluster.local:9000 bucket=tempo`
 5. run `jhack imatrix fill`
 6. wait units settle down (except for metrics-generator). With TLS, it will take around 30 mins to settle
 7. run `juju ssh --container nginx tempo/leader 'nginx -s stop'`. You might want to run that at least 7 times to have some backoff before nginx auto restarts.
 8. run `curl <INGRESS_URL>:3200/api/search` 3 times
 9. You should be able to see the traces in 2/3 curls. 1/3 curls would return a `Bad Gateway` as its routed to the down coordinator instance.

### CODE REVIEW

File: src/charm.py [hunk 1]
```diff
@@ -318,20 +349,35 @@ def _ingress_config(self) -> dict:
...
+                    "loadBalancer": {"servers": self.generate_lb_server_config(self._scheme, port)}
+                }
         return {
             "http": {
                 "routers": http_routers,
                 "services": http_services,
             },
         }
 
+    def generate_lb_server_config(self, scheme: str, port: int) -> List[Dict[str, str]]:
```
PietroPasotti, line 368:
```suggestion
    def _build_lb_server_config(self, scheme: str, port: int) -> List[Dict[str, str]]:
```

for consistency with the other `_build_*_config` methods

src/charm.py
```diff
@@ -97,6 +99,9 @@ def __init__(self, *args):
             ],
         )
 
+        # peer
+        self.framework.observe(self.on[PEER].relation_created, self._on_peer_data_created)
```
PietroPasotti, line 103:
```suggestion
        self.framework.observe(self.on[PEER].relation_created, self._on_peers_relation_created)
```

src/charm.py
```diff
@@ -33,6 +34,8 @@
 from tempo_config import TEMPO_ROLES_CONFIG
 
 logger = logging.getLogger(__name__)
+PEER = "peers"
```
PietroPasotti, line 37:
```suggestion
PEERS_RELATION_ENDPOINT_NAME = "peers"
```

src/charm.py
```diff
@@ -33,6 +34,8 @@
...
+HOSTNAME_KEY = "hostname"
```
PietroPasotti, line 38: this should be included in a pydantic model, not hardcoded in a global

src/charm.py
```diff
@@ -213,6 +229,21 @@ def _on_collect_status(self, e: CollectStatusEvent):
     ###################
     # UTILITY METHODS #
     ###################
+
+    def set_peer_unit_data(self, key: str, data: Any) -> None:
```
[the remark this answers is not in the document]
PietroPasotti: Side-note: given also loki/mimir will probably need something similar, making this information available in the coordinator (next to `Coordinator.peer_addresses`) doesn't seem like a bad idea.

Speaking of, why are we not using `Coordinator.peer_addresses` instead? do we need fqdns for TLS?

PietroPasotti (changes requested):
some concerns about how we're implementing the peers relation data protocol, would like to use a more generic pattern, the rest looks good.

src/charm.py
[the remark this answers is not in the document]
michaeldmitry: I'm not sure I 100% follow.
There should not be any other code modifying peer data. If you mean the other coordinator units, each unit is setting to its own databag, so I'm not sure I understand where would the interfering come from.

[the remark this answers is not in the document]
michaeldmitry: I thought about putting that in the shared object, but this usecase is only valid for Tempo (mimir and loki's IngressPerApp/Unit would automatically do what we want to do with peer data). I thought that one string key is not that error-prone to add a whole databag model for it (inspired by https://github.com/canonical/grafana-k8s-operator/blob/main/src/charm.py#L561 :D). WDYT?

lib/charms/observability_libs/v1/cert_handler.py
```diff
@@ -367,13 +366,15 @@ def __init__(
         )
 
         if refresh_events:
-            for ev in refresh_events:
-                self.framework.observe(ev, self._on_refresh_event)
+            logger.warn(
+                "DEPRECATION WARNING. `refresh_events` is now deprecated. CertHandler will automatically refresh the CSR when necessary."
```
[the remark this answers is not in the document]
michaeldmitry: This and the above comment should be updated in `observability-libs`

src/charm.py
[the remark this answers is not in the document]
michaeldmitry: "Speaking of, why are we not using Coordinator.peer_addresses instead? do we need fqdns for TLS?"
Yes. I need fqdns for TLS and grpc URLs

src/charm.py
```diff
@@ -33,6 +34,14 @@
 from tempo_config import TEMPO_ROLES_CONFIG
 
 logger = logging.getLogger(__name__)
+PEERS_RELATION_ENDPOINT_NAME = "peers"
+
+
+class PeerData(DatabagModel):
+    """Peer data between the coordinator units."""
```
PietroPasotti, line 41:
```suggestion
    """Databag model for the "peers" relation between coordinator units."""
```

src/charm.py
```diff
@@ -33,6 +34,14 @@
...
+
+    hostname: str
+    """FQDN hostname of a coordinator unit."""
```
PietroPasotti, line 44:
```suggestion
    """FQDN hostname of this coordinator unit."""
```

src/charm.py
```diff
@@ -33,6 +34,14 @@
 from tempo_config import TEMPO_ROLES_CONFIG
 
 logger = logging.getLogger(__name__)
+PEERS_RELATION_ENDPOINT_NAME = "peers"
+
+
+class PeerData(DatabagModel):
+    """Peer data between the coordinator units."""
+
+    hostname: str
```
PietroPasotti, line 43: I think we should rename this to "fqdn", so it's more obvious what it means and more future-proof. What if the next time we need to add another type of hostname? what do we call it?

PietroPasotti (approved):
couple more nits, doesn't need another pass from me

PR #62, opened 2024-10-31
Jnsgrukification of tempo-coordinator-k8s

PietroPasotti:


spread.yaml
```diff
@@ -0,0 +1,110 @@
+
+project: zinc-charm-tests
```
mmkay, line 2:
```suggestion
project: tempo-coordinator-k8s-charm-tests
```

spread.yaml
```diff
@@ -0,0 +1,110 @@
...
+      # Set the instance address for spread
+      ADDRESS "$(lxc ls -f csv | grep "${VM_NAME}" | cut -d"," -f3 | cut -d" " -f1)"
+    discard: |
+      instance_name="$(lxc ls -f csv | grep $SPREAD_SYSTEM_ADDRESS | cut -f1 -d",")"
+      lxc delete -f $instance_name
+
+    systems:
+      - ubuntu-24.04:
+          username: spread
+          password: spread
```
mmkay, line 41: Sounds a bit weird to have it in the code.

spread.yaml
```diff
@@ -0,0 +1,110 @@
...
+    systemctl stop unattended-upgrades.service
+    systemctl mask unattended-upgrades.service
+  fi
+
+  snap install --classic concierge
+  concierge prepare --trace -p k8s --extra-snaps astral-uv --extra-debs make
+
+  pushd "$SPREAD_PATH"
+
+  if [[ -f "$PWD/zinc-k8s_amd64.charm" ]]; then
```
mmkay, line 99:
```suggestion
  if [[ -f "$PWD/tempo-coordinator-k8s_amd64.charm" ]]; then
```

spread.yaml
```diff
@@ -0,0 +1,110 @@
...
+    chown $(id -u):$(id -g) $PWD/zinc-k8s_amd64.charm
```
mmkay, line 100:
```suggestion
    chown $(id -u):$(id -g) $PWD/tempo-coordinator-k8s_amd64.charm
```

tests/catan/test_clustering.py
mmkay, line 1: :monocle_face:

tests/integration/conftest.py
```diff
@@ -11,17 +11,135 @@
...
 APP_NAME = "tempo"
+TESTER_NAME = "tester"
+TESTER_GRPC_NAME = "tester-grpc"
+
 SSC = "self-signed-certificates"
 SSC_APP_NAME = "ssc"
 
 logger = logging.getLogger(__name__)
+# Copyright 2021 Canonical Ltd.
+# See LICENSE file for licensing details.
```
mmkay, line 28: Why here and not at the top? Seems like a weird copy-paste.

tests/integration/conftest.py
```diff
@@ -11,17 +11,135 @@
...
+    )
+    return{
+        "workload": meta["resources"]["workload"]["upstream-source"]
+    }
+
+
+@fixture(scope="module")
+def tempo_oci_image():
+    meta = yaml.safe_load(Path("./charmcraft.yaml").read_text())
+    return meta["resources"]["zinc-image"]["upstream-source"]
```
mmkay, line 119:
```suggestion
    return meta["resources"]["tempo-image"]["upstream-source"]
```

mmkay (changes requested):
Bit more renaming needed for it to work.

Makefile
mmkay, line 1: Discussed offline: should there be a `setup uv` makefile action?

.github/workflows/pull-request.yaml
```diff
@@ -1,16 +1,141 @@
...
+          path: ./*.charm
+
+      - name: Pack tester
+        run: charmcraft pack -v -p ./tests/integration/tester/
+
+      - name: Upload charm artifact
+        uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4
+        with:
+          name: tester-charm
+          path: ./tests/integration/tester/*.charm
```
michaeldmitry, line 78: I think the charm will be packed in the same directory as the non-tester one i.e: `./*.charm`

.github/workflows/pull-request.yaml
```diff
@@ -1,16 +1,141 @@
...
+
+      - name: Pack tester-grpc
+        run: charmcraft pack -v -p ./tests/integration/tester-grpc/
+
+      - name: Upload charm artifact
+        uses: actions/upload-artifact@b4b15b8c7c6ac21ea08fcf65892d2ee8f75cf882 # v4
+        with:
+          name: tester-grpc-charm
+          path: ./tests/integration/tester-grpc/*.charm
```
michaeldmitry, line 87: same comment

charmcraft.yaml
```diff
@@ -126,41 +123,72 @@ bases:
...
+      - -.venv
+      - -CONTRIBUTING.md
+      - -Makefile
+      - -pyproject.toml
+      - -README.md
+      - -requirements-dev.txt
+      - -scripts/
+      - -tests/
+      - -uv.lock
+      - -*.egg-info
```
ca-scribner, line 154: Can you explain the significance of this?  I'm not sure why we need it.

Makefile
```diff
@@ -0,0 +1,57 @@
...
+clean:
+	rm -rf .coverage
+	rm -rf .pytest_cache
+	rm -rf .ruff_cache
+	rm -rf .venv
+	rm -rf *.charm
+	rm -rf *.rock
+	rm -rf **/__pycache__
+	rm -rf **/*.egg-info
+	rm -rf requirements*.txt
```
ca-scribner, line 57: Suggest breaking this up, at least to break out the requirements since they're a good thing to regenerate frequently.

```suggestion
clean: clean-charm clean-requirements clean-other

clean-charm:
	rm -f *.charm

clean-requirements:
	rm -f requirements*.txt

clean-other:
	rm -rf .coverage
	rm -rf .pytest_cache
	rm -rf .ruff_cache
	rm -rf .venv
	rm -rf *.rock
	rm -rf **/__pycache__
	rm -rf **/*.egg-info
```

Makefile
```diff
@@ -0,0 +1,57 @@
...
+ALL := $(SRC) $(TESTS)
+
+export PYTHONPATH = $(PROJECT):$(PROJECT)/lib:$(SRC)
+
+update-dependencies:
+	uv lock -U --no-cache
+
+generate-requirements:
+	uv pip compile -q --no-cache pyproject.toml -o requirements.txt
+	uv pip compile -q --no-cache --all-extras pyproject.toml -o requirements-dev.txt
```
ca-scribner, line 14: I dont know when you actually need `requirements-dev.txt` since pip can use pyproject.toml, but I haven't thought about it much
```suggestion
```

Makefile
```diff
@@ -0,0 +1,57 @@
...
+
+lint:
+	uv tool run ruff check $(ALL)
+	uv tool run ruff format --check --diff $(ALL)
+
+fmt:
+	uv tool run ruff check --fix-only $(ALL)
+	uv tool run ruff format $(ALL)
```
ca-scribner, line 22: `uv tool run` does not respect your pyproject.toml or uv.lock files but instead uses the latest tool available.  As is, the `ruff` version called out in pyproject.toml here is meaningless.  To pin a linting tool here, you'd have to hard code it into these commands (`uv tool run ruff@0.7.3` or `uv tool run ruff --from "ruff=0.7.3"`)

An alternative is using `uv run --isolated`, which makes a one-off isolated environment, eg:
```suggestion

lint: EXTRA_DEPENDENCIES = --extra dev
lint:
	uv run --isolated $(EXTRA_DEPENDENCIES) ruff check $(ALL)
	uv run --isolated $(EXTRA_DEPENDENCIES) ruff format --check --diff $(ALL)

fmt: EXTRA_DEPENDENCIES = --extra dev
fmt:
	uv run --isolated $(EXTRA_DEPENDENCIES) ruff check --fix-only $(ALL)
	uv run --isolated $(EXTRA_DEPENDENCIES) ruff format $(ALL)
```

The `EXTRA_DEPENDENCIES = ...` is in there because it was the least objectionable way I saw for handling multiple optional dependency groups (I needed at least 2, because I didn't want pytest operator running during unit tests).  But if having a single group is fine, it can be removed

Makefile
```diff
@@ -0,0 +1,57 @@
...
+	uv run --all-extras \
+		pytest \
+		-v \
+		-x \
+		-s \
+		--tb native \
+		--ignore=$(TESTS)/interface \
+		--ignore=$(TESTS)/unit \
+		--log-cli-level=INFO \
+		$(ARGS)
```
ca-scribner, line 46: afaict this will execute scenario tests.  What about opting in rather than selecting things to exclude?

```suggestion
integration:
	uv run --all-extras \
		pytest \
		-v \
		-x \
		-s \
		--tb native \
		$(TESTS)/integration \
		--log-cli-level=INFO \
		$(ARGS)
```

Makefile
```diff
@@ -0,0 +1,57 @@
...
+fmt:
+	uv tool run ruff check --fix-only $(ALL)
+	uv tool run ruff format $(ALL)
+
+unit:
+	uv run --all-extras \
+		coverage run \
+		--source=$(SRC) \
+		-m pytest \
+		--ignore=$(TESTS)/integration \
```
ca-scribner, line 29: I prefer pointing at the test folders directly (`./tests/unit`, `./tests/scenario`) rather than pointing at the root and then `--ignore`ing things, but both work

Makefile
```diff
@@ -0,0 +1,57 @@
...
+SRC := $(PROJECT)src
+TESTS := $(PROJECT)tests
+ALL := $(SRC) $(TESTS)
+
+export PYTHONPATH = $(PROJECT):$(PROJECT)/lib:$(SRC)
+
+update-dependencies:
+	uv lock -U --no-cache
+
+generate-requirements:
```
ca-scribner, line 12: If you adopt [this suggestion](https://github.com/canonical/tempo-coordinator-k8s-operator/pull/62/files#r1842549141), then:
```suggestion
generate-requirements: clean-requirements
```

I do this because the real `pip-compile` is a bit unpredictable when a requirements.txt already exists, and I don't know how closely `uv pip compile` models the original's behavior.  This feels safer

Makefile
```diff
@@ -0,0 +1,57 @@
...
+		-m pytest \
+		--ignore=$(TESTS)/integration \
+		--tb native \
+		-v \
+		-s \
+		$(ARGS)
+	uv run --all-extras coverage report
+
+integration:
+	uv run --all-extras \
```
ca-scribner, line 37: Jon and I disagree on this, but I think [this should be `uv run --frozen ...`](https://github.com/canonical/parca-k8s-operator/pull/339#discussion_r1757214864).  To me, updating dependencies should be an explicit opt-in rather than a side effect.  The main case for this is debugging an existing charm - if I pull an old charm release and start reproducing an issue, I want `make integration` to use exactly what was released.

To be fair though, `uv run` [often does what a dev would want](https://github.com/canonical/parca-k8s-operator/pull/339#discussion_r1757302817)

Makefile
PietroPasotti: yeah that's a good point, I like that more too

PietroPasotti: however it gets more difficult to work with `$(ARGS)` in that case, because you can't `make unit ./just/this/test.py` anymore, which I think is the reason for the --exclude strategy

Makefile
PietroPasotti: in our case we don't have pytest operator or any 'heavy' dependency in integration tests, so we should be good

Makefile
PietroPasotti: dupe, let's discuss below

charmcraft.yaml
PietroPasotti: it excludes these things from the packed artifact. We don't want to ship our tests in the charm

Makefile
ca-scribner: oh yeah I see what you mean.  

hmm I can't think of a nicer way.  Personally I prefer removing `--ignore` but that's mostly because I never do `make unit ./just/this/test.py`.  But I can see if someone uses it, losing that sucks

PietroPasotti:
some integration tests still fail and take long because of https://github.com/canonical/cos-lib/pull/103

PR #67, opened 2024-11-12
Split workload and charm traces

michaeldmitry:
## Solution
Split `self-tracing` endpoint to `self-charm-tracing` and `self-workload-tracing`

## Context
https://github.com/canonical/cos-lib/pull/95

## Drive-bys
- remove redundant code of pushing nginx CA cert to charm container (handled now by the shared object)


src/charm.py
```diff
@@ -37,6 +37,26 @@
...
+class TempoCoordinator(Coordinator):
+    """A Tempo coordinator class that inherits from the Coordinator class."""
+
+    @property
+    def _charm_tracing_receivers_urls(self) -> Dict[str, str]:
+        """Override with Tempo's enabled and requested receivers."""
+        # if Tempo is related to another Tempo instance, use the endpoints of that instance instead
+        if self.charm_tracing.is_ready():
+            return super()._charm_tracing_receivers_urls
+        return self._charm.requested_receivers_urls()
```
PietroPasotti, line 49: I'd add a comment here to explain that we're returning 'this instance's endpoints' instead

src/charm.py
```diff
@@ -84,14 +104,15 @@ def __init__(self, *args):
...
+                "charm-tracing": "self-charm-tracing",
+                "workload-tracing": "self-workload-tracing",
             },
             nginx_config=NginxConfig(server_name=self.hostname).config,
             workers_config=self.tempo.config,
-            tracing_receivers=self.requested_receivers_urls,
             resources_requests=self.get_resources_requests,
             container_name="charm",
             remote_write_endpoints=self.remote_write_endpoints,  # type: ignore
+            workload_tracing_protocols=["jaeger_thrift_http"],
```
PietroPasotti, line 115: let's add a comment that this will "soon" be replaced by OTEL as soon as they release from main.

src/charm.py
```diff
@@ -298,26 +309,14 @@ def _trace_retention_period_hours(self) -> int:
...
-        else:  # tls unavailable: delete local cert
-            server_ca_cert.unlink(missing_ok=True)
-
     def tempo_otlp_http_endpoint(self) -> Optional[str]:
         """Endpoint at which the charm tracing information will be forwarded."""
         # the charm container and the tempo workload container have apparently the same
         # IP, so we can talk to tempo at localhost.
-        if self.coordinator and self.coordinator.tracing.is_ready():
-            return self.coordinator.tracing.get_endpoint("otlp_http")
+        if self.coordinator and self.coordinator.charm_tracing.is_ready():
```
PietroPasotti, line 318: this is always true, or you get an AttributeError, so if you want to guard against pre-initialization access, we should use 

```
if hasattr(self, "coordinator"): ...
```

PietroPasotti (approved):
couple of nits, but LGTM's!

PR #82, opened 2024-12-02
[DO NOT MERGE] Grafana datasource interface tests

PietroPasotti:
Interface test fixtures for grafana_source interface tests.

See: https://github.com/canonical/charm-relation-interfaces/pull/204

to test: 
`tox -e interface`

Note that the CRI PR should be merged before this one is. At that point we can remove the tester `branch` pin.

### DIFF
--- a/tests/interface/conftest.py
+++ b/tests/interface/conftest.py
@@ -1,5 +1,6 @@
 # Copyright 2022 Canonical Ltd.
 # See LICENSE file for licensing details.
+from contextlib import ExitStack
 from unittest.mock import MagicMock, patch
 
 import pytest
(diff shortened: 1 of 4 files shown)


tests/interface/conftest.py
```diff
@@ -155,3 +155,31 @@ def s3_tester(interface_tester: InterfaceTester):
...
+@pytest.fixture
+def grafana_datasource_tester(interface_tester: InterfaceTester):
+    with patch("lightkube.core.client.GenericSyncClient"):
+        with patch.multiple(
+            "cosl.coordinated_workers.worker.KubernetesComputeResourcesPatch",
+            _namespace="test-namespace",
+            _patch=lambda _: None,
+            is_ready=k8s_resource_patch_ready,
+        ):
+            with patch.multiple(
```
michaeldmitry, line 169: can't this be combined with the above patch?

michaeldmitry: Ok I'm blind

PietroPasotti: copy-pastae'd directly from the scenario tests folder. 
Will give it a shot, but maybe there's a reason they're separate

mmkay: One was in `worker`, one in `coordinator` so let's see if it still works after the removal of one :)

michaeldmitry: No I think its one for the worker and one for the coordinator

Issue #85, opened 2024-12-04
When tempo is deployed on a k8s cluster with CoreDNS instead of kube-dns then nginx fails to start


mmkay:
### Bug Description

As Canonical Kubernetes uses CoreDNS and not kube-dns, nginx fails to start as it cannot find `kube-dns.kube-system.svc.cluster.local` which is hardcoded as a DNS resolver here - https://github.com/canonical/tempo-coordinator-k8s-operator/blob/4633db805340209dce2768bf24b39395e19a780b/src/nginx_config.py#L175

When CoreDNS is there, the resolver is available under `coredns.kube-system.svc.cluster.local`. Was able to confirm this is the case by manually editing nginx config to have `    resolver coredns.kube-system.svc.cluster.local.;` instead. Tempo then starts:

```
tempo/0*         active       idle       10.1.0.207                 metrics-generator disabled. Add a relation over send-remote-write
```

### To Reproduce

Deploy tempo HA on Canonical k8s and cross-relate it.

### Environment

`latest/edge` on Canonical k8s with dns enabled

### Relevant log output

```shell
File "/var/lib/juju/agents/unit-tempo-0/charm/venv/ops/pebble.py", line 2244, in _services_action
    raise ChangeError(change.err, change)
ops.pebble.ChangeError: cannot perform the following tasks:
- Start service "nginx" (cannot start service: exited quickly with code 1)
----- Logs from task 0 -----
2024-12-04T10:56:02Z INFO Most recent service output:
    2024/12/04 10:56:02 [emerg] 22#22: host not found in resolver "kube-dns.kube-system.svc.cluster.local." in /etc/nginx/nginx.conf:44
2024-12-04T10:56:02Z ERROR cannot start service: exited quickly with code 1
-----
unit-tempo-0: 11:56:02 ERROR juju.worker.uniter.operation hook "upgrade-charm" (via hook dispatching script: dispatch) failed: exit status 1
```
```


### Additional context

_No response_

PR #88, opened 2024-12-05
Fix nginx DNS resolution

michaeldmitry:
## Issue
Fixes https://github.com/canonical/tempo-coordinator-k8s-operator/issues/85

- As part of https://github.com/canonical/tempo-coordinator-k8s-operator/pull/66, we added a "dynamic" resolver on each `server` block. However, the parent `http` block had a `resolver` directive that is tied to `kube-dns` while in other K8s flavours, we could be using a different one, such as `coreDNS`
~- Using a `resolver` alone doesn't cover for the scenario where, for example, the worker pod restarts. The entry for worker's fqdn gets evaluated and cached when the nginx config is loaded at the start and doesn't re-resolve when a scenario like this happens, which creates the need to manually restart nginx.~ (will be fixed in a separate PR)

## Solution
~As part of `nginx 1.27.3`, DNS re-resolve and service discovery functionality was pulled from NGINX Plus to the open-source code, where using `resolve` arg in the `upstream` block together with the existing dynamic `resolver` solves the issue.~ (will be added in a diff PR)
Instead of using the hardcoded `kube-dns` resolver, use the dynamic resolver fetched from `/etc/resolv.conf` 

## Testing Instructions
~1- Deploy Tempo HA
2- `curl tempo_ip:3200/api/search`
3- `microk8s.kubectl delete pod -n model tempo-worker-0`
4-  `curl updated_tempo_ip:3200/api/search`
5- you should see some traces there~
1. Install Canonical K8s
2. Deploy Tempo HA
3. `curl tempo_ip:3200/api/search`
4. you should see some traces there


src/nginx_config.py
```diff
@@ -252,7 +241,6 @@ def _build_server_config(
                     {"directive": "ssl_certificate_key", "args": [KEY_PATH]},
                     {"directive": "ssl_protocols", "args": ["TLSv1", "TLSv1.1", "TLSv1.2"]},
                     {"directive": "ssl_ciphers", "args": ["HIGH:!aNULL:!MD5"]},  # codespell:ignore
-                    *self._resolver(custom_resolver=self.dns_IP_address),
```
mmkay, line 255: Not all the uses of  `_resolver` were related to the pod restart. I wonder if we don't break anything with removing it from lines with `custom_resolver` specified.

michaeldmitry: The resolver was pointing to kube-dns. We shouldn't need that since we'd be always deploying this inside a pod in the k8s cluster and we shouldn't have to manually specify a resolver by nature.

+ itests pass :D

PietroPasotti:
the manual test passes for me

charmcraft.yaml
```diff
@@ -26,7 +26,7 @@ resources:
   nginx-image:
     type: oci-image
     description: OCI image for nginx
-    upstream-source: ubuntu/nginx:1.24-24.04_beta
+    upstream-source: nginx:1.27.3
```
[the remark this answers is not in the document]
michaeldmitry: we'll keep the orig image and once we have an ubuntu-provided image/our own rock, we'll have it in a different PR.

michaeldmitry:
Adjusted the scope of this PR to fix deploying Tempo in K8s flavours that don't use `kube-dns` resolver.
Pod IP changes will be tackled as part of this issue.

mmkay (approved):
:heavy_check_mark: Tested to be deploying on Canonical k8s

PR #102, opened 2025-01-14
Stop referring to old and retired library versions

mmkay:
## Issue

There was more leftover docstring updates that I noticed after merging #100. Also we were missing a version bump.

## Solution

Stop referring to old and retired library versions and, where it makes sense, take note of lib migration that happened so an entire section only makes sense for those that were on a very old version. Also we don't expect anyone using `charm_tracing_config` to be on very old tracing library versions anymore.


PietroPasotti (approved):
good catch!

PR #119, opened 2025-02-24
feat: add tempo-api relation

ca-scribner:
## Issue
For other charms to reach the tempo api (for example, to query tempo), we need to expose that URL (https://github.com/canonical/kiali-k8s-operator/issues/26)

## Solution
This adds the tempo_api.py library that implements data accessors for the tempo-api interface, with data model:

```
class TempoApiAppData(BaseModel):
    """Data model for the tempo-api interface."""

    ingress_url: Optional[AnyHttpUrl] = Field(
        default=None,
        description="The non-internal URL at which this application can be reached.  Typically, this is an ingress URL.",
    )
    direct_url: AnyHttpUrl = Field(
        description="The cluster-internal URL at which this application can be reached.  Typically, this is a"
        " Kubernetes FQDN like name.namespace.svc.cluster.local for connecting to the prometheus api"
        " from inside the cluster, with scheme."
    )
```

## Context

https://github.com/canonical/kiali-k8s-operator/issues/26

## Testing Instructions
Scenario tests should cover the lib (provider and requirer) and the usage of the provider in this charm.

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -64,6 +64,10 @@ provides:
     interface: blackbox_exporter_probes
     description: |
       Integration to expose Blackbox Exporter probes for Tempo's reachability.
+  tempo-api:
+    interface: tempo_api
+    description: |
+      Integration to offer other charms the possibility to interact with Tempo's API.
 
 requires:
   self-charm-tracing:
--- a/lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
+++ b/lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
@@ -78,11 +78,12 @@ provides:
 import json
 import logging
 from pathlib import Path
-from typing import Optional
+from typing import Optional, Union
 
 import yaml
 from ops import Application, RelationMapping
-from pydantic import AnyHttpUrl, BaseModel, Field
+from pydantic import AfterValidator, AnyHttpUrl, BaseModel, Field
+from typing_extensions import Annotated
 
 # The unique Charmhub library identifier, never change it
 LIBID = "6d55454c9a104113b2bd01e738dd5f99"
@@ -100,16 +101,20 @@ log = logging.getLogger(__name__)
 
 DEFAULT_RELATION_NAME = "tempo-api"
 
+# Define a custom type that accepts AnyHttpUrl and string, but converts to AnyHttpUrl and raises an exception if the
+# string is not a valid URL
+AnyHttpUrlOrStrUrl = Annotated[Union[AnyHttpUrl, str], AfterValidator(lambda v: AnyHttpUrl(v))]
+
 
 class TempoApiUrls(BaseModel):
     """Data model for urls Tempo offers for query access, for a given protocol."""
 
-    direct_url: AnyHttpUrl = Field(
+    direct_url: AnyHttpUrlOrStrUrl = Field(
         description="The cluster-internal URL at which this application can be reached for a connection."
         " Typically, this is a Kubernetes FQDN like name.namespace.svc.cluster.local for"
         " connecting to the tempo api from inside the cluster, with scheme and maybe port."
     )
-    ingress_url: Optional[AnyHttpUrl] = Field(
+    ingress_url: Optional[AnyHttpUrlOrStrUrl] = Field(
         default=None,
         description="The non-internal URL at which this application can be reached for a connection."
         " Typically, this is an ingress URL.",
@@ -236,10 +241,10 @@ class TempoApiProvider:
 
     def publish(
         self,
-        direct_url_http: AnyHttpUrl,
-        direct_url_grpc: AnyHttpUrl,
-        ingress_url_http: Optional[AnyHttpUrl] = None,
-        ingress_url_grpc: Optional[AnyHttpUrl] = None,
+        direct_url_http: Union[AnyHttpUrl, str],
+        direct_url_grpc: Union[AnyHttpUrl, str],
+        ingress_url_http: Optional[Union[AnyHttpUrl, str]] = None,
+        ingress_url_grpc: Optional[Union[AnyHttpUrl, str]] = None,
     ):
         """Post tempo-api to all related applications.
 
--- a/lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
+++ b/lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
@@ -173,7 +173,9 @@ class TempoApiRequirer:
         if len(relations) == 0:
             return None
 
-        raw_data_dict = relations[0].data.get(relations[0].app)
+        # Being a little cautious here using getattr and get, since some funny things have happened with relation data
+        # in the past.
+        raw_data_dict = getattr(relations[0], "data", {}).get(relations[0].app)
         if not raw_data_dict:
             return None
 
--- a/lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
+++ b/lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
@@ -23,7 +23,7 @@ To use this endpoint wrapper:
 
 * observe the relation-changed event for this relation wherever your charm needs to use this data (this endpoint wrapper DOES
   NOT automatically observe any events)
-* wherever you need access to the data, create a `TempoApiRequirer` instance and use `.get_data()`
+* wherever you need access to the data, call `TempoApiRequirer(...).get_data()`
 
 An example implementation is:
 
--- a/lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
+++ b/lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
@@ -1,0 +1,219 @@
+"""tempo_api.
+
+This library implements data accessors the tempo-api interface.  The tempo-api interface is used to
+transfer information about an instance of Tempo, such as how to access and uniquely identify it.  Typically, this is
+useful for charms that create a Tempo instance and want other applications to be able to access it.
+
+## Usage
+
+### Requirer
+
+Add the following to your `charmcraft.yaml` or `metadata.yaml`:
+
+```yaml
+requires:
+  tempo-api:
+    # The example below uses the API for when limit=1.  If you need to support multiple related applications, remove
+    # this and use the list-based data accessor method.
+    limit: 1
+    interface: tempo_api
+```
+
+To implement handling this relation:
+
+* observe the relation-changed event for this relation wherever your charm needs to use this data (this relation DOES
+  NOT automatically observe any events)
+* wherever you need access to the data, create a `TempoApiRequirer` instance and use `.get_data()`
+
+An example implementation is:
+
+```python
+class FooCharm(CharmBase):
+    def __init__(self, framework):
+        super().__init__(framework)
+        self.framework.observe(self.on["tempo-api"].relation_changed, self._on_tempo_api_changed)
+
+    def do_something_with_metadata(self):
+        # Get exactly one related application's data, raising if more than one is available
+        tempo_api = TempoApiRequirer(self.model.relations, "tempo-api")
+        metadata = tempo_api.get_data()
+        ...
+```
+
+### Provider
+
+Add the following to your `charmcraft.yaml` or `metadata.yaml`:
+
+```yaml
+provides:
+  tempo-api:
+    interface: tempo_api
+```
+
+To manage sending data to related applications in your charm, use `TempoApiProvider`.  Note that
+`TempoApiProvider` *does not* manage any events, but instead provides a `publish` method for sending data to
+all related applications.  Triggering `publish` appropriately is left to the charm author, although generally you want
+to do this at least during relation_joined and leader_elected events.  An example implementation is:
+
+```python
+class FooCharm(CharmBase):
+    def __init__(self, framework):
+        super().__init__(framework)
+        self.tempo_api = TempoApiProvider(
+            relations=self.model.relations,
+            app=self.app,
+            direct_url=self.direct_url,
+            ingress_url=self.external_url,
+            relation_name="tempo-api"
+        )
+
+        self.framework.observe(self.on.leader_elected, self.do_something_to_publish)
+        self.framework.observe(self._charm.on["tempo-api"].relation_joined, self.do_something_to_publish)
+        self.framework.observe(self.on.some_event_that_changes_tempos_url, self.do_something_to_publish)
+```
+"""
+
+import json
+import logging
+from typing import Optional
+
+from ops import Application, RelationMapping
+from pydantic import AnyHttpUrl, BaseModel, Field
+
+# The unique Charmhub library identifier, never change it
+LIBID = "6d55454c9a104113b2bd01e738dd5f99"
+
+# Increment this major API version when introducing breaking changes
+LIBAPI = 0
+
+# Increment this PATCH version before using `charmcraft publish-lib` or reset
+# to 0 if you are raising the major API version
+LIBPATCH = 1
+
+PYDEPS = ["pydantic>=2"]
+
+log = logging.getLogger(__name__)
+
+DEFAULT_RELATION_NAME = "tempo-api"
+
+
+class TempoApiAppData(BaseModel):
+    """Data model for the tempo-api interface."""
+
+    ingress_url: Optional[AnyHttpUrl] = Field(
+        default=None,
+        description="The non-internal URL at which this application can be reached.  Typically, this is an ingress URL.",
+    )
+    direct_url: AnyHttpUrl = Field(
+        description="The cluster-internal URL at which this application can be reached.  Typically, this is a"
+        " Kubernetes FQDN like name.namespace.svc.cluster.local for connecting to the prometheus api"
+        " from inside the cluster, with scheme."
+    )
+
+
+class TempoApiRequirer:
+    """The requirer side of the tempo-api relation."""
+
+    def __init__(
+        self,
+        relations: RelationMapping,
+        relation_name: str = DEFAULT_RELATION_NAME,
+    ) -> None:
+        """Initialize the TempoApiRequirer object.
+
+        This object is for accessing data from relations that use the tempo-api interface.  It **does not**
+        autonomously handle the events associated with that relation.  It is up to the charm using this object to
+        observe those events as they see fit.  Typically, that charm should observe this relation's relation-changed
+        event.
+
+        This object is for interacting with a relation that has limit=1 set in charmcraft.yaml.  In particular, the
+        get_data method will raise if more than one related application is available.
+
+        Args:
+            relations: The RelationMapping of a charm (typically `self.model.relations` from within a charm object).
+            relation_name: The name of the relation.
+        """
+        self._charm_relation_mapping = relations
+        self._relation_name = relation_name
+
+    @property
+    def relations(self):
+        """Return the relation instances for applications related to us on the monitored relation."""
+        return self._charm_relation_mapping.get(self._relation_name, ())
+
+    def get_data(self) -> Optional[BaseModel]:
+        """Return data for at most one related application, raising if more than one is available.
+
+        Useful for charms that always expect exactly one related application.  It is recommended that those charms also
+        set limit=1 for that relation in charmcraft.yaml.  Returns None if no data is available (either because no
+        applications are related to us, or because the related application has not sent data).
+        """
+        relations = self.relations
+        if len(relations) == 0:
+            return None
+        if len(relations) > 1:
+            # TODO: Different exception type?
+            raise ValueError("Cannot get_info when more than one application is related.")
+
+        raw_data_dict = relations[0].data.get(relations[0].app)
+        if not raw_data_dict:
+            return None
+
+        # Static analysis errors saying the keys may not be strings.  Protect against this by converting them.
+        raw_data_dict = {str(k): v for k, v in raw_data_dict.items()}
+
+        return TempoApiAppData.model_validate_json(json.dumps(raw_data_dict))  # type: ignore
+
+
+class TempoApiProvider:
+    """The provider side of the tempo-api relation."""
+
+    def __init__(
+        self,
+        relations: RelationMapping,
+        app: Application,
+        direct_url: AnyHttpUrl,
+        ingress_url: Optional[AnyHttpUrl] = None,
+        relation_name: str = DEFAULT_RELATION_NAME,
+    ):
+        """Initialize the TempoApiProvider object.
+
+        This object is for serializing and sending data to a relation that uses the tempo-api interface - it does
+        not automatically observe any events for that relation.  It is up to the charm using this to call publish when
+        it is appropriate to do so, typically on at least the charm's leader_elected event and this relation's
+        relation_joined event.
+
+        Args:
+            relations: The RelationMapping of a charm (typically `self.model.relations` from within a charm object).
+            app: This application.
+            direct_url: The cluster-internal URL at which this application can be reached.  Typically, this is a
+                        Kubernetes FQDN like name.namespace.svc.cluster.local for connecting to the prometheus api
+                        from inside the cluster, with scheme.
+            ingress_url: The non-internal URL at which this application can be reached.  Typically, this is an ingress
+                         URL.
+            relation_name: The name of the relation.
+        """
+        self._charm_relation_mapping = relations
+        self._data = TempoApiAppData(ingress_url=ingress_url, direct_url=direct_url)
+        self._app = app
+        self._relation_name = relation_name
+
+    @property
+    def relations(self):
+        """Return the applications related to us under the monitored relation."""
+        return self._charm_relation_mapping.get(self._relation_name, ())
+
+    def publish(self):
+        """Post tempo-api to all related applications.
+
+        This method writes to the relation's app data bag, and thus should never be called by a unit that is not the
+        leader otherwise ops will raise an exception.
+        """
+        info_relations = self.relations
+        for relation in info_relations:
+            databag = relation.data[self._app]
+            databag.update(
+                self._data.model_dump(
+                    mode="json", by_alias=True, exclude_defaults=True, round_trip=True
+                )
+            )
--- a/lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
+++ b/lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
@@ -64,7 +64,7 @@ class FooCharm(CharmBase):
         self.framework.observe(self.on.leader_elected, self.do_something_to_publish)
         self.framework.observe(self._charm.on["tempo-api"].relation_joined, self.do_something_to_publish)
         self.framework.observe(self.on.some_event_that_changes_tempos_url, self.do_something_to_publish)
-        
+
     def do_something_to_publish(self, e):
         self.tempo_api.publish(...)
 ```
@@ -102,17 +102,29 @@ log = logging.getLogger(__name__)
 DEFAULT_RELATION_NAME = "tempo-api"
 
 
+class TempoApiUrls(BaseModel):
+    """Data model for urls Tempo offers for query access, for a given protocol."""
+
+    direct_url: AnyHttpUrl = Field(
+        description="The cluster-internal URL at which this application can be reached for a connection."
+        " Typically, this is a Kubernetes FQDN like name.namespace.svc.cluster.local for"
+        " connecting to the tempo api from inside the cluster, with scheme and maybe port."
+    )
+    ingress_url: Optional[AnyHttpUrl] = Field(
+        default=None,
+        description="The non-internal URL at which this application can be reached for a connection."
+        " Typically, this is an ingress URL.",
+    )
+
+
 class TempoApiAppData(BaseModel):
     """Data model for the tempo-api interface."""
 
-    ingress_url: Optional[AnyHttpUrl] = Field(
-        default=None,
-        description="The non-internal URL at which this application can be reached.  Typically, this is an ingress URL.",
+    http: TempoApiUrls = Field(
+        description="The URLs at which this application can be reached for an http connection to query Tempo."
     )
-    direct_url: AnyHttpUrl = Field(
-        description="The cluster-internal URL at which this application can be reached.  Typically, this is a"
-        " Kubernetes FQDN like name.namespace.svc.cluster.local for connecting to the prometheus api"
-        " from inside the cluster, with scheme."
+    grpc: TempoApiUrls = Field(
+        description="The URLs at which this application can be reached for an http connection to query Tempo."
     )
 
 
@@ -166,9 +178,11 @@ class TempoApiRequirer:
             return None
 
         # Static analysis errors saying the keys may not be strings.  Protect against this by converting them.
-        raw_data_dict = {str(k): v for k, v in raw_data_dict.items()}
+        # The data in the databag is in format [str: json-string-of-dict].  Expand out those nested JSON dicts so we
+        # have the full object
+        raw_data_dict = {str(k): json.loads(v) for k, v in raw_data_dict.items()}
 
-        return TempoApiAppData.model_validate_json(json.dumps(raw_data_dict))  # type: ignore
+        return TempoApiAppData.model_validate(raw_data_dict)  # type: ignore
 
     def _validate_relation_metadata(self):
         """Validate that the provided relation has the correct metadata for this endpoint."""
@@ -186,8 +200,6 @@ class TempoApiProvider:
         relations: RelationMapping,
         relation_meta: RelationMeta,
         app: Application,
-        direct_url: AnyHttpUrl,
-        ingress_url: Optional[AnyHttpUrl] = None,
     ):
         """Initialize the TempoApiProvider object.
 
@@ -201,15 +213,9 @@ class TempoApiProvider:
             app: This application.
             relation_meta: The RelationMeta object for this charm relation (typically
                `self.meta.relations[relation_name]`).
-            direct_url: The cluster-internal URL at which this application can be reached.  Typically, this is a
-                        Kubernetes FQDN like name.namespace.svc.cluster.local for connecting to the prometheus api
-                        from inside the cluster, with scheme.
-            ingress_url: The non-internal URL at which this application can be reached.  Typically, this is an ingress
-                         URL.
         """
         self._charm_relation_mapping = relations
         self._relation_meta = relation_meta
-        self._data = TempoApiAppData(ingress_url=ingress_url, direct_url=direct_url)
         self._app = app
         self._relation_name = self._relation_meta.relation_name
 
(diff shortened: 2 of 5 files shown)


lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+```yaml
+requires:
+  tempo-api:
+    # The example below uses the API for when limit=1.  If you need to support multiple related applications, remove
+    # this and use the list-based data accessor method.
+    limit: 1
+    interface: tempo_api
+```
+
+To implement handling this relation:
```
PietroPasotti, line 22:
```suggestion
To use this endpoint wrapper:
```

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+
+```python
+class FooCharm(CharmBase):
+    def __init__(self, framework):
+        super().__init__(framework)
+        self.framework.observe(self.on["tempo-api"].relation_changed, self._on_tempo_api_changed)
+
+    def do_something_with_metadata(self):
+        # Get exactly one related application's data, raising if more than one is available
+        tempo_api = TempoApiRequirer(self.model.relations, "tempo-api")
```
PietroPasotti, line 38: I'd put this in `__init__` and store it in an instance attr, that's where we usually put all endpoint wrappers

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+```yaml
+provides:
+  tempo-api:
+    interface: tempo_api
+```
+
+To manage sending data to related applications in your charm, use `TempoApiProvider`.  Note that
+`TempoApiProvider` *does not* manage any events, but instead provides a `publish` method for sending data to
+all related applications.  Triggering `publish` appropriately is left to the charm author, although generally you want
+to do this at least during relation_joined and leader_elected events.  An example implementation is:
```
PietroPasotti, line 56:
```suggestion
to do this at least during the `relation_joined` and `leader_elected` events.  An example implementation is:
```

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+
+log = logging.getLogger(__name__)
+
+DEFAULT_RELATION_NAME = "tempo-api"
+
+
+class TempoApiAppData(BaseModel):
+    """Data model for the tempo-api interface."""
+
+    ingress_url: Optional[AnyHttpUrl] = Field(
```
PietroPasotti, line 103: how about `ingressed_url`?

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+    )
+    direct_url: AnyHttpUrl = Field(
+        description="The cluster-internal URL at which this application can be reached.  Typically, this is a"
+        " Kubernetes FQDN like name.namespace.svc.cluster.local for connecting to the prometheus api"
+        " from inside the cluster, with scheme."
+    )
+
+
+class TempoApiRequirer:
+    """The requirer side of the tempo-api relation."""
```
PietroPasotti, line 115:
```suggestion
    """Endpoint wrapper for the requirer side of the tempo-api relation."""
```

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+
+        Args:
+            relations: The RelationMapping of a charm (typically `self.model.relations` from within a charm object).
+            relation_name: The name of the relation.
+        """
+        self._charm_relation_mapping = relations
+        self._relation_name = relation_name
+
+    @property
+    def relations(self):
```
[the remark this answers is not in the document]
PietroPasotti: also the line: `self._charm_relation_mapping = relations` is telling in that respect

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+
+        Useful for charms that always expect exactly one related application.  It is recommended that those charms also
+        set limit=1 for that relation in charmcraft.yaml.  Returns None if no data is available (either because no
+        applications are related to us, or because the related application has not sent data).
+        """
+        relations = self.relations
+        if len(relations) == 0:
+            return None
+        if len(relations) > 1:
+            # TODO: Different exception type?
```
PietroPasotti, line 155: if we can prevent this altogether by validating in init, that's best IMO

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+
+    @property
+    def relations(self):
+        """Return the relation instances for applications related to us on the monitored relation."""
+        return self._charm_relation_mapping.get(self._relation_name, ())
+
+    def get_data(self) -> Optional[BaseModel]:
+        """Return data for at most one related application, raising if more than one is available.
+
+        Useful for charms that always expect exactly one related application.  It is recommended that those charms also
```
PietroPasotti, line 147: I don't see the implementation for the limit=N case; what are you supposed to do then?

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+        applications are related to us, or because the related application has not sent data).
+        """
+        relations = self.relations
+        if len(relations) == 0:
+            return None
+        if len(relations) > 1:
+            # TODO: Different exception type?
+            raise ValueError("Cannot get_info when more than one application is related.")
+
+        raw_data_dict = relations[0].data.get(relations[0].app)
```
PietroPasotti, line 158: from experience, many weird things can go wrong here. I'd wrap this in a catch-all, log an exception and return None on failure.

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+        self._data = TempoApiAppData(ingress_url=ingress_url, direct_url=direct_url)
+        self._app = app
+        self._relation_name = relation_name
+
+    @property
+    def relations(self):
+        """Return the applications related to us under the monitored relation."""
+        return self._charm_relation_mapping.get(self._relation_name, ())
+
+    def publish(self):
```
PietroPasotti, line 206: I'd add arguments to this function so that a caller can late-declare ingress url and direct url. Otherwise we may force on our users strange init-order constraints between this object and their ingress wrappers, for example.
So instead of creating the model in `__init__`, create it here in-line. We're not using it anywhere else anyway, so it's also lazier (thus better) to only make it as needed

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+        """Post tempo-api to all related applications.
+
+        This method writes to the relation's app data bag, and thus should never be called by a unit that is not the
+        leader otherwise ops will raise an exception.
+        """
+        info_relations = self.relations
+        for relation in info_relations:
+            databag = relation.data[self._app]
```
PietroPasotti, line 214: also here, there's no such thing as 'too failsafe' when working with relations. 
I'd consider adding this from traefik's ingress; it hasn't let me down yet https://github.com/canonical/traefik-k8s-operator/blob/main/lib/charms/traefik_k8s/v2/ingress.py#L541

src/charm.py
```diff
@@ -304,6 +306,27 @@ def _update_ingress_relation(self) -> None:
...
+        external_url = self._external_url
+        if external_url == internal_url:
+            # external_url is not set and just defaulted back to internal_url.  Set it to None
+            external_url = None
+
+        tempo_api = TempoApiProvider(
+            relations=self.model.relations,
+            ingress_url=external_url,  # pyright: ignore
+            direct_url=internal_url,  # pyright: ignore
+            relation_name=TEMPO_API_RELATION_ENDPOINT_NAME,
```
PietroPasotti, line 325: there's no harm in omitting the default if it matches. A unittest will catch any desync soon enough.

src/charm.py
```diff
@@ -304,6 +306,27 @@ def _update_ingress_relation(self) -> None:
...
+            return
+
+        # tempo-api should only send an external URL if it's set, otherwise it leaves that empty
+        internal_url = self._internal_url
+        external_url = self._external_url
+        if external_url == internal_url:
+            # external_url is not set and just defaulted back to internal_url.  Set it to None
+            external_url = None
+
+        tempo_api = TempoApiProvider(
```
PietroPasotti, line 321: IMO we should strive to keep all relation wrappers inside `__init__`, that's the current design (for better or worse). The fact that we don't need to pass the charm to its constructor doesn't matter: we could do that here as well.

tests/scenario/test_tempo_api_lib.py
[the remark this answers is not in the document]
PietroPasotti: I think an integration test with [a facade charm](https://github.com/canonical/charm-relation-interfaces/pull/152) should be enough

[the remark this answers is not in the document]
PietroPasotti: (so, standalone: I'd rather limit our dependencies on an actual production charms for integration testing)

src/charm.py
```diff
@@ -304,6 +306,27 @@ def _update_ingress_relation(self) -> None:
...
             )
 
+    def _update_tempo_api_relations(self) -> None:
+        """Update all applications related to us via the tempo-api relation."""
+        if not self.unit.is_leader():
+            return
+
+        # tempo-api should only send an external URL if it's set, otherwise it leaves that empty
+        internal_url = self._internal_url
+        external_url = self._external_url
```
mmkay, line 316: `_external_url` will give you the address without the HTTP API port. Is it by choice so that Kiali adds `:3200` by itself?

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+## Usage
+
+### Requirer
+
+Add the following to your `charmcraft.yaml` or `metadata.yaml`:
+
+```yaml
+requires:
+  tempo-api:
+    # The example below uses the API for when limit=1.  If you need to support multiple related applications, remove
```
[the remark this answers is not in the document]
ca-scribner: atm the getter for the no limit case isn't implemented.  Do you think we could change the docs to be clearer with the current code, or should we add that additional getter?

src/charm.py
ca-scribner: I can make this change, but first I'll gently push back :D  

Agreed its a typical pattern to put relation handlers in init, but isn't that because they needed to `framework.observe`?  Putting these in `init` is the cause of this [bend-over-backwards-to-lazily-declare-things](https://github.com/canonical/tempo-coordinator-k8s-operator/pull/119#discussion_r1969334587) situation.  By instantiating `TempoApiProvider` here we eliminate the need to lazily define things

src/charm.py
ca-scribner: 1) I am ignorant to how people interact with Tempo, so I'm sure its wrong and I appreciate your help in fixing it :)
2) should the `external_url` have a port?  I was thinking the ingress_url would be `INGRESS_IP/path-to-tempo`, where port would always be port 80 on the ingress.
3) ...not asked, but... `_internal_url` definitely gives a host without a port, and I think it needs a port?

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: I like this idea with one caveat

If someone instantiates this inside their charm init, they need to handle that error otherwise everything in the charm is blocked.  Does that change your mind on what is best?  I'm not sure how I feel about it.

Alternatively, I don't think this should be instantiated in charm.__init__ at all, but that's a separate thread :D

src/charm.py
mmkay: Yes, so Tempo is peculiar. It doesn't have a UI per se, it has a set of receivers (each under its own port) and API endpoints:
- HTTP API, which is what Grafana / Kiali / others mostly consume, under port 3200
- GRPC API, which is mostly used for automated API operations, under port 9096 (original was 9095, but that's our promtail port)

Because neither of them is the "default" and also GRPC doesn't work well when the application is deployed with a path prefix, we decided to use `traefik-route` and open each of the ports on the ingress instead. So:

- tempo HTTP API is available under `INGRESS_IP:3200`
- tempo GRPC API is available under `INGRESS_IP:9096`
- its OTLP HTTP receiver is under `INGRESS_IP:9318`
...and so on. Please note the lack of `path-to-tempo` here :)

So that's also why both our `external_url` and `internal_url` don't provide a port, because the port you want to use differs based on how you want to interact with Tempo.

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: atm, there is none.  Only the limit=1 case is provided

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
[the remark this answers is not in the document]
ca-scribner: yeah agreed this isn't intuitive.  Part of the problem is that the `model` obj has:
```python
self.model.relations  # is of type RelationMapping
```

should we call `TempoApiProvider.relations` something else?  `relations_list`?  Or, should we change the Provider to `TempoApiProvider(relation_mapping=self.model.relations,...`?

I think I like changing the provider arg better

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,219 @@
...
+        event.
+
+        This object is for interacting with a relation that has limit=1 set in charmcraft.yaml.  In particular, the
+        get_data method will raise if more than one related application is available.
+
+        Args:
+            relations: The RelationMapping of a charm (typically `self.model.relations` from within a charm object).
+            relation_name: The name of the relation.
+        """
+        self._charm_relation_mapping = relations
```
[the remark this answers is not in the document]
ca-scribner: that is a perfectly workable solution that I sorta hate :D 

That check is needed exactly once (whenever a dev changes the charmcraft.yaml file), but if implemented here it'll be tested every time the charm is executed in the field forever.  

This situation feels like a unit test at most is needed, but we can't require a unit test on a user of this lib... so then we're stuck either just documenting "use this with limit=1" like done here, or adding the check you describe.  I lean toward documenting, but if you prefer the check we can do that

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: related thread: https://github.com/canonical/tempo-coordinator-k8s-operator/pull/119/files/b66e87feef79e6a11efdf26b9a317a755b428a31#r1969343473

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: related: https://github.com/canonical/tempo-coordinator-k8s-operator/pull/119/files/b66e87feef79e6a11efdf26b9a317a755b428a31#r1969343473

tests/scenario/test_tempo_api_lib.py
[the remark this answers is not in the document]
ca-scribner: Do you have an example of what a facade for the requirer should look like here?  I see an example in `present_facade` that sends data, but I'm not sure how this helps if I'm mocking the requirer side.

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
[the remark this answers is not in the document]
ca-scribner: Fleshing this out a bit more...  it would schematically look something like

```python
def validate_relation_limit():
  charm_root = Path(__file__).absolute().parent.parent.parent.parent.parent
  try:
    relation = yaml.load(charm_root / "charmcraft.yaml")["requires"][relation_name]
  except (KeyError, FileNotFoundError): 
    # Can't find the relation or charmcraft.yaml, try metadata.yaml
    relation = yaml.load(charm_root / "metadata.yaml")["requires"][relation_name]

  limit = relation.get("limit", None)
  if limit != 1:
    raise Exception(...)
```

That works, but only be escaping the encapsulation using knowledge about charms.  I don't think as a user of this wrapper, I expect it to read my filesystem

What about passing the `RelationMeta` (`self.meta[relation_name]`) instead of `relation_name`?  ex:

```python
class TempoApiRequirer:
    def __init__(
        self,
        relation_meta: RelationMeta
        relations: RelationMapping,
    ) -> None:
        self.relation_meta = relation_meta
        self.relation_name = relation_meta.relation_name  # Not needed, just a convenience
        self.relations = relations
        self.validate_relation_configuration()

    def validate_relation_configuration():
        if self.relation_meta.limit != 1:
            raise Exception(...)
```

I still feel its a unit check sitting in user space, but its no worse than checking `if len(relations) > 1` and now it doesn't escape encapsulation.  wdyt?

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: Thinking on this more, I still have reservations about this sort of "can fail intermittently at runtime" check in init.  I have a specific example from istio's ingress for why I think its important, but we can discuss offline if interested.

Regardless though, I think this'll get resolved by [this thread](https://github.com/canonical/tempo-coordinator-k8s-operator/pull/119#discussion_r1969323368) if we enforce limit=1

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: Agreed that weird things can happen here, but not sure we should hide those from the users of the Requirer.

What sort of failures are you thinking?  Like that it couldn't get data even though the data exists?  Those sound like good cases for an uncaught exception that should trigger a retry of the event.  Otherwise, we're stuck in this broken state until an unrelated event fires.  Maybe I'm not thinking of the same failures.  Can you elaborate?

The traefik ingress [is_ready](https://github.com/canonical/tempo-coordinator-k8s-operator/pull/119#discussion_r1969330835) example is a different case I think.  That's not masking a `get_data` error, that's *using* a `get_data` error to make a decision about readiness.  A better comparison I think would be if traefik had this:

```python
    def get_data(self, relation: Relation) -> IngressRequirerData:
        """Fetch the remote (requirer) app and units' databags."""
        try:
            return IngressRequirerData(
                self._get_requirer_app_data(relation), self._get_requirer_units_data(relation)
            )
        except (pydantic.ValidationError, DataValidationError) as e:
            # MODIFIED
            # raise DataValidationError("failed to validate ingress requirer data") from e
            return None
```

but then its harder for the caller of `get_data` to know what went wrong.  Was the data empty?  Was it a validation error?

src/charm.py
ca-scribner: After discussing offline, we are changing the object such that the constructor only takes invariate configurations (eg: relations, relation_name, and app) and then `publish` always requires the data to be published.  Eg:

```python
tempo_api = TempoApiProvider(
    relations=self.model.relations,
    relation_name=TEMPO_API_RELATION_ENDPOINT_NAME,
    app=self.app,
)

tempo_api.publish(urls...)
```

Then we'll put this constructor in the charm's init, but there's no worry about lazy declaration of things

charmcraft.yaml
```diff
@@ -60,6 +60,10 @@ provides:
     interface: tracing
     description: |
       Integration to offer other charms the possibility to send traces to Tempo.
+  tempo-api:
```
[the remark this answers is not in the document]
ca-scribner: We decided to have a data model that is a mapping of {protocol: all_urls_for_that_protocol}, eg:

```
http:
  ingress_url: http://abc.com
  direct_url: http://...
grpc:
  ...
```

src/charm.py
ca-scribner: notes from a discussion on this [here](https://github.com/canonical/tempo-coordinator-k8s-operator/pull/119/files/b66e87feef79e6a11efdf26b9a317a755b428a31#r1971944928)

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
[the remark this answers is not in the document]
ca-scribner: Docs are updated to make this clearer.  wdyt?

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: done

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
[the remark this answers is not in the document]
ca-scribner: Added this in the PR.  See `_validate_relation_metadata` in the TempoApiRequirer.  Thoughts?

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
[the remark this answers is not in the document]
ca-scribner: Renamed the arg for init to `relation_mapping`.  wdyt?

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: This should be clearer in the docstrings so I'm resolving it.  There's another open thread about this same thing so if that has an issue I'll reopen this one too

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: I think this is resolved now with the `_validate_relation_metadata` method.  I'll close this if the other thread gets a :+1:

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: Updated the api to accept the data here, not in init

src/charm.py
ca-scribner: This is updated and I think we now send the correct urls/ports

src/charm.py
ca-scribner: Done

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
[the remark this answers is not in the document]
PietroPasotti: huh. Didn't know the 'limit' field was accessible through RelationMeta! This works. 
The price we pay for it is make the API harder to use. It's a tradeoff, but I get that escaping the box may feel strange.
On the other hand, we're not writing to the filesystem, only reading a metadata file, so IMHO it's not that bad a thing to do.

Instead of the chain of `.parent`, FYI, we could also do:

`Path(os.environ["CHARM_DIR"])/"metadata.yaml"`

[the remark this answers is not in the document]
PietroPasotti: So in summary, I don't have a strong opinion either way, but IMHO a lean UX in return for escaping the box and reading a metadata file is a fair tradeoff, so I'd go with that.

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
PietroPasotti: agreed, it's not a case of 'failing intermittently at runtime', but a case of 'this will 100% fail all the time unless you do it right, and you should catch this in every unittest in which you are covering this branch'. So the only case in which a charm without limit=1 will make it into live juju is one in which the user isn't writing (good) unittests, so that's on them.

And to be fair, if they don't write unittests, how likely are they to be reading the docs?

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
PietroPasotti: IIRC `relation.data` might be None, `relation.data[smth]` might also be None.
Agreed that what traefik is doing is different, but it is super-defensive and that pays off.
The goal is for the model to be eventually consistent, so in that light we don't always care why there is no data, only that right now there is no data, and next time there may be.
if you want the hook to fail so juju can retry it, that's also a valid point, but we need to document this well so the caller can make a choice. In the case of tempo, for example, I'd not want the charm to go in error state

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,223 @@
...
+An example implementation is:
+
+```python
+class FooCharm(CharmBase):
+    def __init__(self, framework):
+        super().__init__(framework)
+        self.framework.observe(self.on["tempo-api"].relation_changed, self._on_tempo_api_changed)
+
+    def do_something_with_metadata(self):
+        # Get exactly one related application's data, raising if more than one is available
```
PietroPasotti, line 37:
```suggestion
        # Get the related Tempo's API data
```

it should be obvious by now that there can/should only be at most 1

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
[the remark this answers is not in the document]
ca-scribner: sorry, I forgot to push the changes last night when writing this.  It should be clearer now

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: Yep this was a leftover.  Removed

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,223 @@
...
+            return None
+        if len(relations) > 1:
+            # TODO: Different exception type?
+            raise ValueError("Cannot get_info when more than one application is related.")
+
+        raw_data_dict = relations[0].data.get(relations[0].app)
+        if not raw_data_dict:
+            return None
+
+        # Static analysis errors saying the keys may not be strings.  Protect against this by converting them.
```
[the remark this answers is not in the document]
ca-scribner: The typing issue actually doesn't exist anymore thanks to some refactoring.  Removing this conversion

src/charm.py
```diff
@@ -304,6 +306,27 @@ def _update_ingress_relation(self) -> None:
...
+        if external_url == internal_url:
```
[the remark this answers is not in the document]
ca-scribner: `TempoApiProvider` doesn't have logic to know if the ingress is a real ingress or not, that's the charm's responsibility.

I'm happy to refactor, but the reason that it isn't just `if self.ingress.is_ready()` is because that's [not what tempo does](https://github.com/canonical/tempo-coordinator-k8s-operator/blob/c3fd559a8e9685b6cad9ee9c513fddc636c6ea8a/src/charm.py#L195).  But we could refactor to share the check from `_external_url` if you'd rather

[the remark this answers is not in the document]
dstathis: Can they ever be the same legitimately?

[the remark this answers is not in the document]
ca-scribner: I don't think so, though maybe there's a strange edge case?

Its a good point though that using `if external_url == internal_url` as a test for "do i have an external url" feels, at best, confusing for a future developer.  so imo refactoring somehow sounds smart

Thinking about this more broadly, to me the confusion here is that `self._external_url` **can return a non-external url**.  Its more like `self.._most_external_url`.  What about refactoring such that:
* `self._external_url`: returns the external_url, if available, or None
* adding `self._most_external_url` (or some other name): always returns a url, defaulting back to the internal if needed

That way if the external_url returns None if its not available

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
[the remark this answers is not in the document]
simskij: I think Pietro's comment makes sense. Please add a notice about it in the top.

tests/scenario/test_tempo_api_lib.py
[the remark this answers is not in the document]
PietroPasotti: fine with unittests only for now.

src/charm.py
PietroPasotti: @mmkay ya happy?

src/charm.py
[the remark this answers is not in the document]
PietroPasotti: deferred to separate PR

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+class TempoApiAppData(BaseModel):
+    """Data model for the tempo-api interface."""
+
+    http: TempoApiUrls = Field(
+        description="The URLs at which this application can be reached for an http connection to query Tempo."
+    )
+    grpc: TempoApiUrls = Field(
+        description="The URLs at which this application can be reached for an http connection to query Tempo."
```
PietroPasotti, line 126:
```suggestion
        description="The URLs at which this application can be reached for a GRPC connection to query Tempo."
```

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,276 @@
...
+            app: This application.
+            relation_name: The name of the wrapped relation.
+        """
+        self._charm_relation_mapping = relation_mapping
+        self._app = app
+        self._relation_name = relation_name
+
+    @property
+    def relations(self):
+        """Return the applications related to us under the monitored relation."""
```
PietroPasotti, line 234:
```suggestion
        """Return the relations currently alive on this endpoint."""
```

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
```diff
@@ -0,0 +1,276 @@
...
+                              object).
+            app: This application.
+            relation_name: The name of the wrapped relation.
+        """
+        self._charm_relation_mapping = relation_mapping
+        self._app = app
+        self._relation_name = relation_name
+
+    @property
+    def relations(self):
```
PietroPasotti, line 233:
```suggestion
    def relations(self) -> Tuple[ops.Relation, ...]:
```
?

src/charm.py
```diff
@@ -333,6 +343,31 @@ def _update_ingress_relation(self) -> None:
...
+        external_url = self._external_url
+        if external_url:
+            ingress_url_http = external_url + f":{self.tempo.server_ports['tempo_http']}"
+            ingress_url_grpc = external_url + f":{self.tempo.server_ports['tempo_grpc']}"
+        else:
+            ingress_url_http = None
+            ingress_url_grpc = None
+
+        self.tempo_api.publish(
+            direct_url_http=direct_url_http,  # pyright: ignore
```
PietroPasotti, line 365: do we need these `: ignore`s?

tests/scenario/test_tempo_api_implementation.py
```diff
@@ -0,0 +1,162 @@
...
+            ),
+            "grpc": json.dumps(
+                {
+                    "direct_url": TEMPO_COORDINATOR_URL + f":{Tempo.server_ports['tempo_grpc']}/",
+                }
+            ),
+        }
+
+    # Assert
+    assert tempo_api.local_app_data == expected
```
PietroPasotti, line 77: uhm I don't think this test is doing what you think it's doing.

`tempo_api` is immutable, and is the relation as you defined it before the event is processed.
What you need is:

```
state_out = context.run(context.on.relation_joined(tempo_api), state=state)
tempo_api_out = state_out.get_relation(tempo_api.id)
assert tempo_api.local_app_data == expected
```

tests/scenario/test_tempo_api_lib.py
```diff
@@ -0,0 +1,179 @@
...
+    with tempo_api_provider_context(
+        # construct a charm using an event that won't trigger anything here
+        tempo_api_provider_context.on.update_status(),
+        state=state,
+    ) as manager:
+        manager.charm.relation_provider.publish(**sample_data_to_tempo_api_publish_args(data))
+
+    # Assert
+    # Convert local_app_data to TempoApiAppData for comparison
+    actual = TempoApiAppData.model_validate(
```
PietroPasotti, line 116: same here

PietroPasotti (changes requested):
a few docs suggestions left, the rest looks good!
The only thing we really need to adjust are the unittests. At the moment they're quite broken.

tests/scenario/test_tempo_api_implementation.py
ca-scribner: Turns out it did work, and [not on purpose](https://github.com/canonical/operator/issues/1605) :)  Refactored as suggested.

tests/scenario/test_tempo_api_lib.py
ca-scribner: fixed!

src/charm.py
ca-scribner: full disclosure: I'm not 100% sure how to work with pydantic models directly

This was a fun one.  I think I've got it sorted by doing this to the data model:

```python
AnyHttpUrlOrStrUrl = Annotated[Union[AnyHttpUrl, str], AfterValidator(lambda v: AnyHttpUrl(v))]

class TempoApiUrls(BaseModel):
    """Data model for urls Tempo offers for query access, for a given protocol."""

    direct_url: AnyHttpUrlOrStrUrl = Field(
        description="The cluster-internal URL at which this application can be reached for a connection."
        " Typically, this is a Kubernetes FQDN like name.namespace.svc.cluster.local for"
        " connecting to the tempo api from inside the cluster, with scheme and maybe port."
    )
    ingress_url: Optional[AnyHttpUrlOrStrUrl] = Field(
        default=None,
        description="The non-internal URL at which this application can be reached for a connection."
        " Typically, this is an ingress URL.",
    )
```

That means the model accepts `AnyHttpUrl` and `str`, but it'll still validate the str using `AnyHttpUrl`.  The reason we need this new type is that we can't simply use `[AnyHttpUrl, str]` because `TempoApiUrls(direct_url='definitely-not-a-url')` would have been acceptable to the data because its a valid string (even though its not a url)

lib/charms/tempo_coordinator_k8s/v0/tempo_api.py
ca-scribner: Used `def relations(self) -> List[Relation, ...]:` instead because the RelationMapping returns a list, but otherwise :+1:

src/charm.py
PietroPasotti: ok I don't get it but I'm sure it's needed. fine by me!

PietroPasotti (approved):
thanks for the patience!

PR #122, opened 2025-02-27
refactor: change TempoCoordinatorCharm._external_url, add ._most_external_url

ca-scribner:
## Issue
Previously, TempoCoordinatorCharm._external_url returned the external url if it was available, otherwise it returned the ingress url.  This meant that any consuming code that explicitly wanted the external_url or nothing needed to duplicate checks that exist in external_url, making it easy to later have these go out of sync.

## Solution
This change breaks _external_url into two functions:
* _external_url, which explicitly returns only an external url or None (if it isn't available)
* _most_external_url, which returns the external_url if it is available, otherwise the internal url

This way the failover behaviour of the previous external_url is preserved, but we can also access exactly the external_url when needed.

## Context
#119's need for knowing exactly the external_url is the driver for this change

## Testing Instructions
Existing CI should cover this.  All that changes here is an intermediate product is also now available

## Upgrade Notes
No effects

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -94,14 +94,14 @@ class TempoCoordinatorCharm(CharmBase):
         # set the open ports for this unit
         self.unit.set_ports(*self.tempo.all_ports.values())
 
-        self.tracing = TracingEndpointProvider(self, external_url=self._external_url)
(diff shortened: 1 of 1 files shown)


PietroPasotti (approved):
lgtm!

PR #129, opened 2025-03-06
Redirect Non-TLS traffic through ingress when TLS is enabled

### LINKED ISSUE #116: Non-TLS traffic through ingress shouldn't be allowed when TLS is enabled
### Bug Description

Currently, when TLS is enabled, users can still hit the the ingress endpoint using `http` and Traefik charm would allow it.

### To Reproduce

1. Deploy Traefik
2. Deploy any app (e.g: Tempo)
3. Integrate Traefik and Tempo over `traefik_route`
4. Deploy SSC
5. Integrate Traefik and SSC
6. Curl any valid Tempo endpoint using Traefik's ingress URL but use `http://<endpoint>` not `https://<endpoint>`

### Environment

microk8s 1.28.13
juju 3.4.5

### Relevant log output

```shell
N/A
```

### Additional context

_No response_

michaeldmitry:
## Issue
Fixes #116 

## Solution
Add a redirect middleware in traefik_route config for every port tempo exposes.

## Testing Instructions
Run `test_tls.py` itest

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -452,6 +452,7 @@ class TempoCoordinatorCharm(CharmBase):
                 if self.ingress.is_ready() and self.ingress.scheme == "https"
                 else {}
             )
+            middlewares.update(redirect_middleware)
 
             http_routers[f"juju-{self.model.name}-{self.model.app.name}-{sanitized_protocol}"] = {
                 "entryPoints": [sanitized_protocol],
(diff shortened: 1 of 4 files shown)


PietroPasotti (approved):
looks good overall, a couple of nits, but it doesn't need another pass from me

PR #131, opened 2025-03-11
Test blocked state when ingress is over TLS but Tempo is not

michaeldmitry:
Fix itests for the changes coming in https://github.com/canonical/cos-lib/pull/133


PietroPasotti (approved):
should be good to go now without the git+ pins

PR #138, opened 2025-03-17
tolerate existing apps in deploy_minio script

PietroPasotti:
Updated the deploy_minio script to not fail if the s3-integrator and minio applications already exist. 
Useful for configuring already existing deployments (e.g. created by a bundle)


scripts/deploy_minio.py
```diff
@@ -59,18 +62,26 @@ def deploy(
...
+            _run(f"juju deploy {ch_name} --channel edge --trust {app}")
+        else:
+            print(f"found app {app} already deployed.")
 
-    _run(
-        f"juju deploy minio --channel edge --trust --config access-key={user} --config secret-key={password} {minio_app_name}"
-    )
-    _run(f"juju deploy s3-integrator --channel edge --trust {s3_app_name}")
+    print("configuring minio credentials...")
+    _run(f"juju config {minio_app_name} access-key={user} secret-key={password}")
```
mmkay, line 82: This might be tricky in case we have another minio (for other purposes) under the expected name in this model. We then change its access key and secret key, potentially breaking the previous application that used it. 

Maybe we could, instead, try to deploy a new minio instance (with a suffix in the name) if one already exists?

cos-lite_r11.bundle
mmkay, line 1: Leftover, needs to go?

PietroPasotti:
Stale. Reopening in #178

PR #144, opened 2025-04-16
feat: uvfy repo and bump to CI v1

PietroPasotti:
This PR adds an uv lock, changes tox to use uv instead of direct commands,
separates the dependencies of the various testing environment in dependency groups and uses them from tox
bumps all workflows to use luca's CI V1

## Testing Instructions
In principle if the CI passes we're good


charmcraft.yaml
lucabello, line 1: Please make sure to follow [this](https://github.com/canonical/observability-team/blob/main/_temporary/how-to-onboard-charms-to-ci-v1.md#modify-charmcraftyaml-again) to the letter. You need to change the `bases` key into the newer `platforms` one, otherwise the new CI won't spin up runners correctly.

lucabello: You should also, in general, take the opportunity to put `optional: true`  to most relations - tempo-coordinator and worker are an exception because they need at least the other one and s3 to do something

pyproject.toml
```diff
@@ -8,22 +8,49 @@ requires-python = ">=3.10"
...
     "pydantic<3",
     # ---PYDEPS---
     # lib/charms/tempo_coordinator_k8s/v0/charm_tracing.py
     "opentelemetry-exporter-otlp-proto-http==1.21.0",
 ]
 
-[project.optional-dependencies]
-dev = [
+
+[dependency-groups]
```
lucabello, line 21: We use only one dependency group named `dev`, not one per type of test

tests/integration/helpers.py
```diff
@@ -302,9 +302,9 @@ def tempo_worker_charm_and_channel_and_resources():
     """Tempo worker charm used for integration testing.
 
     Build once per session and reuse it in all integration tests to save some minutes/hours.
-    You can also set `TEMPO_WORKER_CHARM` env variable to use an already existing built charm.
+    You can also set `WORKER_CHARM_PATH` env variable to use an already existing built charm.
     """
-    if path_from_env := os.getenv("TEMPO_WORKER_CHARM"):
+    if path_from_env := os.getenv("WORKER_CHARM_PATH"):
```
lucabello, line 307: Nothing in CI is setting this, though I'm not sure what your worker CI does

tox.ini
lucabello, line 1: You can remove `CHARM_BUILD_DIR` and `MODEL_SETTINGS` from `passenv`; the `tox` file should use the `dev` group for all dev dependencies like other charms do

lucabello (changes requested):

tests/integration/helpers.py
PietroPasotti: it's just for local testing

charmcraft.yaml
```diff
@@ -130,24 +145,21 @@ peers:
...
-        channel: "22.04"
+platforms:
+  ubuntu@22.04:amd64:
 
 parts:
   charm:
     # uncomment this if you add git+ dependencies in requirements.txt:
     # build-packages:
     #  - "git"
-    charm-binary-python-packages:
```
michaeldmitry, line 146: We don't need the above comment anymore

tests/catan/test_clustering.py
```diff
@@ -92,9 +92,9 @@ def tempo_coordinator():
     tempo = App.from_path(
         REPOS_ROOT / "tempo-coordinator-k8s-operator",
         patches=[
-            patch("charm.KubernetesServicePatch"),
+            # patch("charm.KubernetesServicePatch"),
             patch("lightkube.core.client.GenericSyncClient"),
-            patch("charm.TempoCoordinatorCharm._update_server_cert"),
+            # patch("charm.TempoCoordinatorCharm._update_server_cert"),
```
michaeldmitry, line 97: Do we need to keep those commented?

tests/integration/helpers.py
```diff
@@ -283,28 +102,21 @@ async def deploy_and_configure_minio(ops_test: OpsTest):
...
 
 def tempo_worker_charm_and_channel_and_resources():
     """Tempo worker charm used for integration testing.
 
     Build once per session and reuse it in all integration tests to save some minutes/hours.
-    You can also set `TEMPO_WORKER_CHARM` env variable to use an already existing built charm.
+    You can also set `WORKER_CHARM_PATH` env variable to use an already existing built charm.
     """
-    if path_from_env := os.getenv("TEMPO_WORKER_CHARM"):
+    if path_from_env := os.getenv("WORKER_CHARM_PATH"):
```
michaeldmitry, line 119: We should update the worker's CI with this new env name

michaeldmitry (commented):
- Added some minor comments
- We should also update the `tiobe-scan` workflow file

charmcraft.yaml
PietroPasotti: uhm, why not? I'll give it a try

tests/catan/test_clustering.py
PietroPasotti: yeh well I temporarily turned off catan tests as I couldn't get them to work with newer ops. Adding a ticket to track.

tests/integration/helpers.py
PietroPasotti: yes, I'm not sure the worker CI will work anyway given the v1 migration we're doing here. we'll need to fix it soon regardless.

charmcraft.yaml
michaeldmitry: because you've already added that in line 158

PietroPasotti: :facepalm:

tests/catan/test_clustering.py
PietroPasotti: https://github.com/canonical/tempo-coordinator-k8s-operator/issues/145

will add references in comments too

mmkay (approved):

PR #148, opened 2025-04-25
Feature/use pytest jubilant

PietroPasotti:
Use pytest-jubilant for integration tests.

### DIFF
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -1,10 +1,9 @@
 import json
-import shlex
-import subprocess
 from pathlib import Path
 
 import jubilant
 import pytest
+from pytest_jubilant.main import pack_charm
 import yaml
 from jubilant import Juju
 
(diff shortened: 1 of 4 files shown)


tests/integration/test_integration.py
```diff
@@ -19,40 +18,26 @@
...
-
 @pytest.mark.setup
 def test_build_deploy_tester(juju: Juju):
-    tester_root = "./tests/integration/tester/"
-    proc = subprocess.run(shlex.split("charmcraft pack -p %s" % tester_root),
-                          check=True, capture_output=True, text=True)
-    tester_charm = proc.stderr.strip().splitlines()[-1].split()[-1]
-    resources_tester = {"workload": TESTER_METADATA["resources"]["workload"]["upstream-source"]}
-
+    out = pack_charm("./tests/integration/tester/")
```
michaeldmitry, line 23: maybe we should use this in `_get_tempo_charm` as well?

PietroPasotti: uhm, good idea

PR #155, opened 2025-05-01
Simplify nginx config generation

michaeldmitry:
## Issue
All our current three coordinators generate their own nginx config from scratch although they share a lot of common config.

## Solution
Extract all shared configuration into `cosl`, and keep only the coordinator-specific parts of the `upstreams` and `locations` blocks, which each coordinator will pass to the shared object as needed.

## Context
https://github.com/canonical/cos-lib/pull/143

## Drive-bys
- Update lock file


src/nginx_config.py
```diff
@@ -3,344 +3,67 @@
...
 
 from tempo import Tempo
 from tempo_config import TempoRole
 
 logger = logging.getLogger(__name__)
-RESOLV_CONF_PATH = "/etc/resolv.conf"
 
 
-class NginxConfig:
+class NginxHelper:
```
PietroPasotti, line 23: I'd get rid of this class altogether. all it has left is a couple of staticmethods: let's lift them to globals or toplevel functions and be done with it.

PietroPasotti (approved):
nice! that's what I like to see :)

PR #163, opened 2025-05-21
use k8s service as internal url

### LINKED ISSUE #143: use the charm's `service` as the internal_url, not a headless service pointing to the leader pod
### Enhancement Proposal

When the charm needs to communicate its [`internal_url`](https://github.com/canonical/tempo-coordinator-k8s-operator/blob/ad90b9c14dded5aaba04942c3c7345afa47cc8a5/src/charm.py#L250-L252) (the url to reach the charm inside the cluster), it does a [`socket.getfqdn()`](https://github.com/canonical/tempo-coordinator-k8s-operator/blob/ad90b9c14dded5aaba04942c3c7345afa47cc8a5/src/charm.py#L205) which returns something like `http://tempo-coordinator-k8s-0.tempo-coordinator-k8s-endpoints.cos.svc.cluster.local` (`POD_NAME.DEPLOYMENT_NAME.NAMESPACE.svc.cluster.local`).  This url is a [headless service](https://kubernetes.io/docs/concepts/services-networking/service/#headless-services) - basically a service that **points to a single pod** (in this case, the leader).  This means that all traffic always hits the tempo-coordinator leader, regardless of the number of units.

afaict, tempo-coordinator is a stateless frontend for the workers of tempo, so it doesn't matter which tempo-coordinator unit is accessed.  In that case what I think you want to be sending as the `internal_url` is the applications regular service created by juju (`tempo-coordinator-k8s.cos.svc.cluster.local`, or `SERVICE_NAME.NAMESPACE.svc.cluster.local`).  This will loadbalance across all the units

A nice side effect of doing this is that it'll also fix https://github.com/canonical/istio-k8s-operator/issues/30 :)

michaeldmitry:
Fixes #143 

let the coordinator advertise its `internal_url` as the K8s service (which will by nature loadbalance traffic across all coordinator units.)

## Testing Instructions
Itests should cover testing this change. IMO, we shouldn't be testing that K8s services actually do their job in loadbalancing.

### DIFF
--- a/tests/integration/test_integration.py
+++ b/tests/integration/test_integration.py
@@ -45,6 +45,15 @@ def test_deploy_monolithic_cluster(juju: Juju, tempo_charm: Path):
     # Then applications should eventually be created
     deploy_monolithic_cluster(juju)
 
+# scaling the coordinator before ingesting traces to verify that scaling won't stop traces ingestion.
+def test_scale_up_tempo(juju: Juju):
+    # GIVEN we scale up tempo
+    juju.add_unit(TEMPO_APP, num_units=2)
+    # THEN all units become active
+    juju.wait(
+        lambda status: jubilant.all_active(status, TEMPO_APP, WORKER_APP, TESTER_APP_NAME, TESTER_GRPC_APP_NAME),
+        timeout=1000
+    )
 
 @pytest.mark.setup
 def test_relate(juju: Juju):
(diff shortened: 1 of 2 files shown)


tests/integration/test_integration.py
```diff
@@ -45,6 +45,15 @@ def test_deploy_monolithic_cluster(juju: Juju, tempo_charm: Path):
     # Then applications should eventually be created
     deploy_monolithic_cluster(juju)
 
+# scaling the coordinator before ingesting traces to verify that scaling won't stop traces ingestion.
```
michaeldmitry, line 48: mark as `@setup`

